PR #54, opened 2019-09-11
Removes extra header element

gcardonag:


gcardonag:
Overriding approval for hotfix

PR #66, opened 2019-09-24
change the accordian in ClosestTap to just a text changer

jlap4567:


gcardonag (approved):
Awesome, way better use of our valuable space 👍

PR #70, opened 2019-11-06
Adds graceful handling of scenario where user location is not available

gcardonag:


gcardonag:
Merging this to get the bugfix out

WaltRibeiro:
🙌

PR #75, opened 2020-01-26
Adds basic Jest tests and Lighthouse testing

gcardonag:
Also includes test reporting, which is done by publishing to our AWS account. Test results for all commits can be reviewed at https://test.phlask.me

The current approach generates a test site for each commit pushed (or for the last commit if more than one is pushed in a set). These test sites last for one day, which should be enough time to review them.


gcardonag:
Talked it over during the weekly meetup and we agreed to merge it in. Merging.

PR #76, opened 2020-01-28
implements search bar with autocomplete suggestions and map recentering

PeterArriaza:
- Added react-places-autocomplete dependency for pre built React search bar with autocomplete.  
- Search bar is an input with type="search" to give native browser clear input functionality, which is only HTML5 compliant.  
- Removes search bar from Head.js and overlays it on Map.  
- Input width is 300px so should fit on all mobile devices.
- adds "center" prop to <Map> in ReactGoogleMap which is set to state variables currlat & currlon to implement map recentering on address search


gcardonag (approved):

gcardonag:
Ignore the failed checks as the site was tested locally and confirmed to work. I'll retest on the develop branch.

PR #77, opened 2020-02-16
Adds Redux and implements map filtering 

PeterArriaza:


gcardonag:
I only just started playing with it on the test site but for some reason enabling the filtered flag only shows Private-Shared Taps. Also a few things that would be nice to look into (I can help with this but wanted to document them somewhere before going to rest):

- It would be better if the entire map isn't moved when a tap is selected. As a user I'd feel pretty frustrated if I was panning out to some other place but the map would redraw and recenter to my current location whenever I tried to filter taps at that location.
- Previously, we had it set up such that clicking the tap icon would grey the icon out in the legend and remove taps from the map, and vice-versa. That made it easier to show the user what filters were active vs which are inactive. It would be nice to have the filters work that way, if that's possible in the react version of the site.

PeterArriaza:
I briefly looked through some of the data points and it seemed that only
the private-shared taps are listed as having filtered water.  I didn't do
an exhaustive double check of this though.  A lot of the data is listed as
"unsure" which I treated as "no" and so they get filtered out too.

I can definitely change the color of the taps instead of removing them.  I
thought that I'd show a different version of the filtering process because
I feel like the gray-ing out process keeps the map a little too busy.
Maybe we can disable the details card from popping up when a user clicks
the grayed out tap?

Let's talk more tonight!

gcardonag:
D’oh! I thought filtered was a flag to show/hide actively filtered tap types. The wording there tripped me up. Not sure if there’s a way to improve that down the line... However, the functionality is a-ok by me then.

I wont be able to join the meet tonight due to a late work event but I would say it would be nice to use that meet to show the functionality to the rest of the team and see what the consensus is for how filtering should feel.

waltribeiro:
In reference to the map moving when a tap is selected - perhaps a drawer
appearing would be a better option.

Attached picture - this would allow users to scroll endlessly, should our
data sheets grow in the future.

gcardonag (approved):
This is a solid start to filtering. I'll open a separate branch to take a crack at making filtering occur without redrawing the map. I think that'll require enough work to merit its own separate PR.

PR #79, opened 2020-02-19
new map markers and fixed typo in SearchBar.js

waltribeiro:


gcardonag (approved):
These new tap icons are sweet 🎉

Issue #97, opened 2020-04-14
Identify Nearest Tap


gcardonag:
As a user, I need the PHLASK Button to register my location and identify the nearest Tap

PR #101, opened 2020-04-25
Brazill/selected tap info

Brazill-Mitchell:
In **ReactGoogleMaps.js**, I moved the filtered tap list to `componentDidMount `& `componentDidUpdate`, so they would only re-render when the filter list changes. I also added a check for finding the `closestTap`, so that it's not called on every update. There are issues with these approach which I have listed below.

**Changes**

- Moved filtered tap list out of render area
- Added a check for Closest Tap
- Moved API key to own file
- Moved firabase.init to index.js
- Grouped Filter Component with Selected Tap

**Known Issues (Things I'm working on but you're welcome to step in)** 

- Location Pin doesn't go away after clicking a new marker
- Clicking the selected marker a second time makes info disappear (not getting correct props)

**Fixes Needed**

- Closest tap is only updated if there is no closest tap in the State. We'll need a function to check if the user's location has changed [significantly enough] to update it. 
- Currently, `setTimeout ` is being used in `componentDidMount ` to delay the mapping `filteredTaps` until some other tasks have completed. Since I'm not too familiar with what the other tasks are, I hackily used `setTimeout `to let them complete so `filteredTaps ` would get the props they need 

**Note**

For now the UI has only been designed for mobile until a desktop layout is concrete.


src/components/ReactGoogleMaps.js
```diff
@@ -175,11 +202,45 @@ export class ReactGoogleMaps extends Component {
         this.setState({ currlon: position.coords.longitude });
       }
     });
+
+    setTimeout(() => {
```
Brazill-Mitchell, line 206: `setTimeout `should be changed to fire after filteredTaps have loaded

src/components/ReactGoogleMaps.js
```diff
@@ -241,18 +308,16 @@ export class ReactGoogleMaps extends Component {
   render() {
     console.log(this.props.filterFunction);
     if (this.props.allTaps.length) {
-      var closestTap = closest(this.props.allTaps, {
-        lat: this.state.currlat,
-        lon: this.state.currlon
-      });
+      if((Object.keys(this.state.closestTap).length) === 0){
```
Brazill-Mitchell, line 310: Instead of checking for the length of `this.state.closestTap`, should check if location have changed enough to change the closest tap

gcardonag: This is OK for now, the functionality to trigger relocating the closestTap can be left to building the closest tap button here: https://github.com/phlask/phlask-map/issues/97

gcardonag (approved):
🎉

PR #102, opened 2020-05-03
Fixes redrawing map on filters

gcardonag:
- Moves filtering to Redux state
- Adds selectors to implement filter-based tap selecting


gcardonag:
Showed this on our meeting today (May 5) to @Brazill-Mitchell and @rah0x2C8 and got the go-ahead from @PeterArriaza to merge this in. Once the filtering branch is merged in on top of this there will be some work expected to connect the new TapMarkers component to the state to interface with the Tap Modal

PR #103, opened 2020-05-13
Brazill/selected tap info

Brazill-Mitchell:
Moved some more states to Redux, and added a couple of properties to MapMarkers. 
Removed Search, Closest tap, and Filter from render area for now since they will be moved to the Toolbar. 

**Note: most features only work in mobile view until I update the interface for desktop, which will start now.


gcardonag:
The Selected Tap component is looking good! I made a small change to fix it being rendered invisibly and blocking the map on desktop for some reason.

Also, is the Filter meant to be hidden? It seems that the old Filter was removed but the new one is tied to a Toolbar component that isn't being rendered.

gcardonag (approved):
Looks good! 👍

PR #106, opened 2020-06-07
Brazill/selected tap info

### LINKED ISSUE #84: Food/Water Toggle
As a user, I need two buttons to toggle on / off the Food OR Water interfaces

Brazill-Mitchell:
This PR adds the food data to the project, a toggle button to switch between the two views, as well as some updates to the desktop view.
The toggle button is just an icon located on the right side, which indicates the mode it will switch to when clicked.

**Issues**

- Currently, the Marker components make a call to the database each time they're mounted (switching between views). We can have the components get their info from redux by default, except the first render since it would be empty.

- I'm working on an issue I noticed with the hours. They don't display properly when the org is only open a couple of days.


src/components/ReactGoogleMaps.js
```diff
@@ -244,16 +256,39 @@ export class ReactGoogleMaps extends Component {
...
+
+          {/* FilteredTaps */}
+
+          {this.props.phlaskType === PHLASK_TYPE_WATER
+            ? <WaterFilter/>
+            : <FoodFilter/>
+          }
 
-            <Filter/>
+          {/* Issue: MapMarkers won't render when placed inside container? */}
```
gcardonag, line 268: Is this still an issue? Looks like the markers are all rendering properly in the test site

gcardonag:
Overall this looks good! The issue with the tap/food toggle repulling all data will be an issue as our list of data grows but shouldn't be a huge burder as we are now (each DB call for all tap/food data is currently at <200KB each).

I wasn't able to review the issue with the hours in time for the test site to expire but overall it seemed ok with the markers I saw. I left a comment regarding the issue listed in the comments to see if that's still needed but otherwise we can chat about this on tomorrow's meeting and I can approve (or I can approve sooner/later depending on your reply to that comment)

src/components/ReactGoogleMaps.js
Brazill-Mitchell: For some reason when I place the Marker components inside a parent element to conditionally render with with their respective filters, the filters render but the markers don't. They show up in devtools but not on the map.
It's not a breaking issue, I solved it by rendering the Markers separate from their filter (ideally they'd be together). It was mostly just a comment for me to come back to later haha

PR #115, opened 2020-07-16
Brazill/selected tap info

Brazill-Mitchell:
### + Toolbar
This pull adds the toolbar with functionality.
**This version doesn't get the tap's icon when choosing nearest tap.
**It also has a sizing issue when viewed from the a phone


gcardonag:
The page appears to load great on desktop and mobile. The mobile sizing issue has been recorded in https://github.com/phlask/phlask-map/issues/116

gcardonag (approved):
🎉

PR #118, opened 2020-07-31
svg marker and filter icons

LizKalter:
Replace the png format marker and filter icons with a pair of js functions that return the icons in URI encoded svg format. The color and size of the icons is set via the function parameters.


gcardonag:
Closing this PR since we merged the other one that included these changes: https://github.com/phlask/phlask-map/pull/119

PR #119, opened 2020-08-01
Update logo

LizKalter:
Replace the main logo with updated versions for standard and retina display so that the logo won't look blurry.


gcardonag (approved):

gcardonag:
Already talked about it over our meetup but this is awesome! I love storing the SVGs in code and being able to select the color dynamically with hex codes 😄

PR #124, opened 2020-09-02
Fixed position for mobile toolbar

### LINKED ISSUE #116: Toolbar not immediately accessible on mobile view
**Describe the bug**
The bottom toolbar for the map does not appear on the initial view of the app on a mobile view, because the app's height appears to be slightly larger than the visible screen. This requires a user to scroll down to get to the toolbar, which can be difficult since the map itself is also scrollable.

**To Reproduce**
1. Go to beta.phlask.me on a phone
2. See if the bottom toolbar is visible.
3. If it is not, try to scroll down the page (Note: This can be done by holding to scroll from the top menu bar)

**Expected behavior**
The bottom toolbar should be visible at all times when on the mobile view of the webapp.

**Screenshots**
N/A (this can be reproduced by going to beta.phlask.me)

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
iPhone on Safari

**Additional context**

LizKalter:
Fixes issue #116


gcardonag (approved):
Looking good on iPhone 8 & 11 as well as my desktop and laptop! 🎉

PR #126, opened 2020-09-07
Replace toolbar icons with svgs

LizKalter:
These are the new svg food and water icons for the toolbar that were set up during the 9/1 hack night.


gcardonag (approved):
The new icons look crisp on desktop and my iPhone! 🎨

PR #129, opened 2020-09-15
Rasha/modal

RashaMou:
This feature adds a popup tutorial modal whenever the app launches, and an info icon to the toolbar which launches the modal at any time.


LizKalter:
On the mobile version, there is an alignment issue with the info icon in the toolbar. I would try moving your new button out of the "filterButton" div so that it can be positioned relative to the toolbar. Give the button itself a class name to apply the positioning to and use right: 2%. You'll also have to adjust the positioning on the button for desktop after making this change. Feel free to contact me with any questions.

RashaMou:
@LizKalter I fixed the button position per your suggestion and added some additional styling enhancements.

LizKalter:
I think your changes have inadvertently affected the filter popup for both desktop and mobile. That's the popup you get if you click on the menu icon in the toolbar. Also the search box doesn't look quite right either. Please take a look at https://beta.phlask.me/ for comparison. I think you may have changed some global styles that are affecting other areas of the site.

RashaMou:
Oh, I see. I think the way I added custom styling to React Bootstrap components is overriding some other styles. I'll push up a fix.

RashaMou:
Fixed!

LizKalter:
Looks good to me, thank you! I have requested that @gcardonag do the official review for approving this PR.

gcardonag (approved):
Really nice, I like the use of the icons to call out to elements of the app! Looks great on mobile and desktop 👍

PR #133, opened 2020-09-30
Center map above mobile modal

LizKalter:


gcardonag (approved):
Worked a-ok on my phone! The tap icon gets covered by some info modals if the tap has too much info and the map is zoomed in too much but it only affected one out of many taps I checked so it might not be a big issue.

PR #134, opened 2020-09-30
Liz/refactor hours

LizKalter:
This is prep work for switching to the new food data feed. Added a default food filter icon in case none of the "access" values matches a named switch case. Moved all of the logic for displaying hours out of the SelectedTap component and into a new SelectedTapHours component.


gcardonag (approved):
Good idea! Looks good on my phone and desktop 👍

PR #135, opened 2020-09-30
Add map zoom controls for desktop

LizKalter:
Added the zoom controls for desktop, which necessitated changing a few things around. Additional comments below.


src/components/Filter.module.scss
```diff
@@ -93,6 +93,7 @@
 .desktopPopover {
   width: 30vw;
   min-width: 400px;
+  max-width: none;
```
LizKalter, line 96: Found another width issue with the filter popup.

src/components/Head.css
```diff
@@ -27,11 +27,6 @@
   height: 66px;
 }
 
-/* style the wrapper div around the map that we don't seem to be able to assign a class name to  */
-.mobile-menu + div {
-  height: 100%;
-}
-
```
LizKalter, line 34: Don't need some of this mobile specific stuff anymore. Instead I added a new wrapper div around the parent map object.

src/components/ReactGoogleMaps.js
```diff
@@ -2,7 +2,7 @@ import { Map, GoogleApiWrapper, Marker } from "google-maps-react";
 import React, { Component } from "react";
 import ReactTouchEvents from 'react-touch-events';
 import SearchBar from "./SearchBar";
-import "./ReactGoogleMaps.css";
+import styles from "./ReactGoogleMaps.module.scss";
```
LizKalter, line 5: Converted a couple of the css files to modules.

src/components/ReactGoogleMaps.js
```diff
@@ -268,6 +268,11 @@ export class ReactGoogleMaps extends Component {
                 className={"map"}
                 style={style}
                 zoom={this.state.zoom}
+                zoomControl={!isMobile}
+                streetViewControl={false}
+                mapTypeControl={false}
+                rotateControl={false}
+                fullscreenControl={false}
```
LizKalter, line 275: There were some css rules that I removed that were hiding all of the map controls. Instead these props will hide them. The zoom control will only be visible for desktop. Also, there is some Google branding that was being hidden by the css rules that is now visible at the bottom of the page. I believe that technically we are required to display these things as part of the terms of service.

src/components/MapPage.js
```diff
@@ -46,14 +46,16 @@ export class MapPage extends Component {
...
-      </div>
+        <div className={styles.mapPage}>
+          <Head legendButton={this.legendButton} />
+          <div className={styles.mapWrapper}>
+            <ReactGoogleMaps
+              tapsDisplayed={this.state.tapsDisplayed}
+              ada={this.state.ada}
+              filtered={this.state.filtered}
+            />
+          </div>
```
LizKalter, line 55: The map was getting pushed down off the page and clipping the zoom controls, so this new wrapper div allows the map to be properly sized and positioned below the header. An extra wrapper div gets added automatically around the map without any class name, making it inconvenient to style, hence the need for this additional wrapper div with a class applied to it. There are issues opened in the google-maps-react project referencing this issue and there seems to be no good resolution for it.

gcardonag: Since the app header was moved off into the App component in develop, I made a small change since I was getting a double-header in the test site.

gcardonag (approved):
Thanks for the detailed comments! 😄 Tested it out on desktop and mobile with the test site and both look good 👍

PR #141, opened 2020-10-18
Updates to tap modal

LizKalter:
This PR replaces the fountain photo on the tap modal with the tessellation artwork.


gcardonag (approved):
Really liking the tessellation placeholder! 🎨

PR #142, opened 2020-10-20
Add checkbox to dismiss tutorial modal on load

RashaMou:
The modal now has a checkbox that says 'don't show me this again' that sets the user preference in local storage using a useLocalStorage hook


LizKalter:
@gcardonag @RashaMou PR #136 relocates all logic for displaying the tutorial modal from the Toolbar component into the TutorialModal component itself. The button was relocated to the upper right of the map to make room for the icon for displaying the Add Tap modal. Merging this PR and that one will require some additional work to resolve conflicts.

RashaMou:
@LizKalter @gcardonag Relocating the logic makes sense. I can convert this PR to a draft and when PR #136 is merged, I'll address any conflicts before submitting for review.

LizKalter:
@RashaMou @gcardonag They might decide to merge these changes in before #136. I just wanted to make sure everyone was aware of some pending merge conflicts between the two updates, to allow extra time for looking into it. If I can be of any help let me know.

gcardonag:
This looks good! My only concern is that clicking the dismiss option immediately closes the modal, which may not be what is expected from selecting the checkbox. My thinking is that if a person wants to see the tutorial but not see it after that time, the checkbox serves to guarantee only that, while closing the modal itself can be left to the close button instead.

To do this, we could remove the effect hook and instead set the default state for showModal to the output of showModalPreference. That way on the initial render, the modal will show but, if the user clicks the checkbox, the showModal will default to false due to the updated showModalReference upon the next render.

What do you think @RashaMou ? I'm ok with approving the merge request as-is, but wanted to throw this idea in beforehand.

RashaMou:
Yup you’re right, I didn’t think of that! I’ll make the change tonight and you can merge with the new behavior.

RashaMou:
@gcardonag checkbox behavior changed as discussed.

gcardonag:
Thanks @RashaMou ! I can confirm the behavior worked as expected on the test site. 👍

gcardonag (approved):

PR #143, opened 2020-11-23
Fixes tap image upload functionality using signed URLs

gcardonag:
Changed the upload approach to use POST requests via signed URLs instead of using the S3 library. The main benefit with this approach is that we can enforce the number of images loaded and unique filenames. I also made a change so that we have a different bucket and DB target for test vs beta vs production. That way we can test freely in each environment without interfering with the other environments.


LizKalter (approved):
Looks good from my end! Maybe we want to add a confirmation message after the form is submitted, but that can be a separate PR.

PR #144, opened 2020-12-01
Pete/tap distances

PeterArriaza:
selectedTaps now show walking duration times.  This has minimal styling and may need to be modified.  Uses fetch request in componentDidUpdate lifecycle to [openRouteService API ](https://openrouteservice.org/) using my own hardcoded credentials.  This will allow us to create requests for directions  at 2.000 requests per day @ 40 requests per minute.


package.json
```diff
@@ -12,7 +12,7 @@
     "geolocation": "^0.2.0",
     "google-maps-react": "^2.0.2",
     "jquery": "^3.4.1",
-    "node-sass": "^4.14.1",
+    "node-sass": "4.9.0",
```
LizKalter, line 18: Not sure why this is downgraded?

package.json
```diff
@@ -7,13 +7,17 @@
...
+    "@types/googlemaps": "^3.40.3",
+    "@types/markerclustererplus": "^2.1.33",
+    "@types/react": "^17.0.0",
     "bootstrap": "^4.3.1",
     "firebase": "^6.2.0",
     "geolocation": "^0.2.0",
     "google-maps-react": "^2.0.2",
     "jquery": "^3.4.1",
-    "node-sass": "^4.14.1",
+    "node-sass": "4.9.0",
```
LizKalter, line 18: I think this may be why the build is failing. 4.9 is an older version.

PR #145, opened 2020-12-03
Pete/tap distances and times

PeterArriaza:
added fetch request in SelectedTap.js to [openrouteservice API](https://openrouteservice.org/) using hardcoded API key to get distance in miles and time from user's location to selected tap


gcardonag (approved):
Looks great on the test site!🚶

PR #148, opened 2020-12-09
Modal hours display

### LINKED ISSUE #146: Food location hours inconsistently disappear
**Describe the bug**
Upon loading the map and reviewing the People’s Emergency Center food location at 3750 Lancaster Ave., the hours do not appear. However, upon clicking another location nearby (for example, Westpark Apartments at 300 N. Busti Street) and then clicking back upon the original location, the hours show up again. It is not clear what is causing this behavior, as no errors are shown in the console.

**To Reproduce**
1. Load the beta map (beta.phlask.me)
2. Switch to food taps
3. Click on the People’s Emergency Center food location at 3750 Lancaster Ave.
4. Hours do not appear

**Expected behavior**
An hours dropdown should appear at the location

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

LizKalter:


src/components/SelectedTapHours.js
```diff
@@ -14,7 +14,7 @@ const SelectedTapHours = ({ infoIsExpanded, selectedPlace }) => {
 
   useEffect(() => {
     // update hoursList, currentOrgHours, and isOpen when selectedPlace changes
-    if (selectedPlace.hours !== undefined) {
+    if (selectedPlace.hours && selectedPlace.hours.length === 7) {
```
LizKalter, line 17: This will ensure that the hour related data is null if we don't have all 7 entries in the dataset. "Closed" will be displayed without a dropdown.

src/components/SelectedTapHours.js
```diff
@@ -14,7 +14,7 @@ const SelectedTapHours = ({ infoIsExpanded, selectedPlace }) => {
...
       const hoursList = [];
 
       selectedPlace.hours.map((orgHours, index) => {
```
LizKalter, line 20: `selectedPlace.hours.map((orgHours, index) => {`

This section expects hours to have 7 entries. The days of the week are 0 indexed with 0 as Sunday. It looks like most of the data for the food taps do not have 7 entries. Most of them have 5.

LizKalter:
Doesn't look like I can comment on just any line in the file that I want to. This is in reference to lines 46-69 beginning with:

`if (selectedPlace.hours[currentDay] !== undefined) {`

This section runs if the current day's index is in the dataset. The weird behavior we are currently seeing is because "People’s Emergency Center food location at 3750 Lancaster Ave." only has 2 entries (index 0 and 1) and we are viewing this on a Tuesday which is index 2, so this code block does not run. If this is your first time opening the modal, currentOrgHours will remain null. The hours dropdown is not displayed when currentOrgHours is null.

Then if you go to a different tap that has a longer dataset for the hours so that this block of code runs, it will set the currentOrgHours value, which is never reset when you go back to the "People’s Emergency Center" location. Hence you see the dropdown.

keegan2149:
I figured it out but I can't push to this branch.  The dates are being processed twice.  Once in `selectedPlace.hours.map((orgHours, index) => {}` and again in `hours.js`.  The `null` meant to trigger the conditional at the end of `SelectedTapHours.js` was being fed to `hours.js` which tried to convert it to INT and returned `NaN`.  I added a conditional at the end of `hours.js` to return `null` when `NaN` is detected.

```
diff --git a/src/components/hours.js b/src/components/hours.js
index dc6ff1e..d1e53d7 100644
--- a/src/components/hours.js
+++ b/src/components/hours.js
@@ -64,10 +64,13 @@ export const hours = {
         //     ? `${newHour}${mer}`
         //     : `${newHour}:${minute}${mer}` );
         
-    
-        return parseInt(minute) === 0
+        if (isNaN(hour) || isNaN(minute) ){
+            return null
+        } else {
+            return parseInt(minute) === 0
             ? `${newHour}${mer}`
             : `${newHour}:${minute}${mer}`    
+        }
     },
     

```

gcardonag (approved):
Looks great! 3750 Lancaster Ave tap shows that its closed on Tuesdays properly now 👍

PR #166, opened 2021-02-10
updated json with cypress commands

zeeatwork:
I have some working tests, but testing the tap selection is tricky.  Our test passes if we enter the xy coordinates for our window size, but the test fails if we resize the window.  I'm researching a way to return a consistent result in the Cypress docs.


zeeatwork:
I just wanted some feedback on these build failures.  I've only changed the Cypress testing files in the codebase.  I also mistakingly installed NPM/Node locally.  Let me know if there anything that I need to repair to complete a successful merge.

gcardonag:
@zeeatwork I haven't been able to fully review the tests yet but its nice that they all seem to work on my machine and on the fixed GitLab build process! I had to make some small changes to account for parameters I was passing for Jest testing that aren't compatible with Cypress. Those changes are here: https://github.com/phlask/phlask-map/pull/166/commits/663cc47d1876f8170af4b8c425f5c57938546ba5

I also cleared out the npm lockfile which doesn't seem to affect the build 🎉

gcardonag (approved):
Hooray for more testing! 🛠️

PR #168, opened 2021-02-24
Keegan/crowdsourcing

keegan2149:
Updated add tap forms
Separate component for food taps
Dropdowns for elegance and to save space
Moved image upload to the top so it is more noticable


keegan2149:
I renamed some of the firebase fields to make the form more palatable for the users. Can we update the firestore attributes? How big is the datastore?  If it's viable I'll write the code.

gcardonag:
Closing this one in favor of #200 , thanks for your work in preparing this Keegan!

PR #170, opened 2021-03-05
Enhancement: Phlask Project Page Content

nwyin:
Hi!

I reached out to @affabillyty and was directed to a shared `Site` folder on Google Drive that had content from the old site. I thought that a small improvement to this version of the application would be migrating some of the content from the old version to this one.

If you want a direct link the folder/document I've referred to with this content, please shoot me a DM on the CodeForPhilly Slack. My username on there is `tbn33`

Thanks for taking the time to review this PR!

image

_Edit: I thought it would be helpful to also include a screen shot of the updated page_


gcardonag (approved):
✏️

PR #187, opened 2021-05-31
Fixes Find Closest Tap button bug

gcardonag:
Fixes a user-reported bug where using the "Find Closest Tap" button would break the site.

The root cause turned out to be a result of the fact that setting the currently selected tap (via the SET_SELECTED_PLACE redux action) can be done by two approaches, passing in an ID or an object. When we pass an ID, we then store an object reference to a location relative to our client-side copy of the tap/food DBs. When we pass an object, the object structure can be anything that is passed over.

After reviewing references to the setSelectedPlace() function, when we pass an object, we are passing a reference to the Marker element we are using to store the tap details. In that element, we store the lat/lng under a position property, which is what the getWalkingDurationAndTimes() function in the SelectedTap was originally looking for. In order to address this short term to fix functionality, we are simply adding the `lat` and `lon` properties to the Food/Water Marker objects and searching for those lat/lon properties, which will exist for both the Marker objects as well as DB items.

### DIFF
--- a/src/components/IndieFoodMarker.js
+++ b/src/components/IndieFoodMarker.js
@@ -148,6 +148,11 @@ class IndieMarker extends React.Component{
                   position={{ lat: this.props.org.lat, lng: this.props.org.lon }}
                   icon={this.getIcon(this.props.org.access)}
                   infoIcon={this.getIcon(this.props.org.access,true)}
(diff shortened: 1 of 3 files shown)


gcardonag:
Overriding approval requirement since this is a hotfix.

Issue #199, opened 2021-09-14
Write unit tests for our crowdsourcing forms


gcardonag:
**Is your feature request related to a problem? Please describe.**
As a PHLASK developer, I would like to ensure changes to the crowdsourcing components do not affect the ability for users to successfully submit taps of any type.

**Describe the solution you'd like**
In order to verify this for all commits in our repository, we would like to add a new test called `crowdsourcing-form.js` under this location: https://github.com/phlask/phlask-map/tree/develop/cypress/integration 

For guidance on how to write the test, see:
- The official guide from Cypress: https://docs.cypress.io/guides/getting-started/writing-your-first-test
- One of our previous tests: https://github.com/phlask/phlask-map/blob/develop/cypress/integration/water-tap-viewing.spec.js

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

mandyshri101:
Any idea how can I run the cypress run command here ? I am trying to run the existing cypress tests by running cypress run command. It is saying this error: Could not start Xvfb. Install XVFB and run Cypress again

gcardonag:
Replacing this ticket with the following ones that break this task out a bit more #467 #468 #469 #470 #471 #472 
@mandyshri101 feel free to take any of these to continue developing crowdsourcing tests

Issue #196, opened 2021-06-16
Implement Google address auto-complete in form


katjost:
Currently, it is painful to manually enter a full address into the crowdsource form.

Ideally, Google maps would auto-complete the address / show suggested addresses based on what is being inputted.

nwyin:
Working on this and have a version working!

Could use some touch up on stylings + handling edge cases, but works for some common use cases I tested for. If we could some more hands to test inputs and just verify that it works would probably give me more peace of mind.

![image](https://user-images.githubusercontent.com/19523341/133502646-8d345a2d-f387-46b9-a43e-09051ada07d7.png)

tagging @affabillyty to keep him in the loop.

nwyin:
Closing this for now given that crowdsourcing form issue has been closed and PR is merged. Address auto-complete is working, but admittedly it's limited based on Google Map's API.

Future enhancements to this feature should be created in a new issue.

PR #200, opened 2021-09-16
Feature: Crowdsourcing Forms

### LINKED ISSUE #113: Crowdsourcing tap information
Ability for in-app native crowdsourcing of PHLASK tap data

PHLASK tap data crowdsourcing would fall into two categories:
1. New PHLASK taps (food & water)
2. Existing PHLASK tap information

**New PHLASK Taps:** Users should be able to submit new taps and any corresponding information they have for a tap. 
**Existing PHLASK Taps:** Users can submit information that is missing from a tap OR propose changes to existing information

Crowdsourced Information:
Users should be able to submit crowdsourced information for any PHLASK Tap's data fields

Verification Process
Crowdsourced information is submitted to a separate database and manually reviewed before being approved and merged into the primary PHLASK Tap database

nwyin:
Hello! I am so excited to finally get this PR in. Super sorry it took ~2 months to develop everything (and there's still more we could do), but now I feel like there have been enough substantial changes and minimum functionality that it makes sense to get this reviewed and tested by more people than just myself.

I've tried my best to break down and summarize all the work that's been done on this branch (along with my best attempt at good commit messages), but I know that it is confusing to look at other people's work. Please reach out to me at @tbn33 on Slack if there are any questions or concerns.

## Summary of Changes
**TL;DR: Crowdsourcing forms are up and functional, and are ready for testing on the `test` environment**

-  `AddTapModal.js` has been changed to `AddResouceModal.js`, and all related components, styles, and utility logic are grouped together in `src/components/AddResourceModal`
    - It's my recommendation that as components grow more complex we move to this structure for new and existing components to make it easier for new (and existing) developers to identify the different domains and functions of the app
- `AddTapModal.js` has additional state and event handlers to accommodate new resource types. Specifically, the "view" of the modal is controlled by a new state attribute called `formStep`, which is modified to show the different forms for different resources.
- To try and reduce copy + pasted code, resource forms have been split into their own sub-components, which are composed of the `ShareFormFields.jsx` component and then additional custom fields specific to that resource
- Unrelated to the crowdsourcing form, by the `cypress.json` file has been modified such that none of the example tests are run. I've made this change as the example tests are non-relevant to the Phlask repo and make running tests take more time. 
- I've created additional test databases on Firebase for foraging and bathroom resources, title `phlask-web-map-test-foraging-verify` and `phlask-web-map-test-bathroom-verify` respectively.
- Address suggestion/autocomplete on each crowdsourcing form is functional. This relates to issue #196.

## Known Issues + Areas of Improvement
- `firebase` is subject to throwing errors as a result of having multiple database connections open at the same time. Specifically, when using the React Dev Tools add-on on Chrome, an error is thrown when using the `Profiler` tab to try and read state. I cannot reproduce the bug otherwise, so for the time being I am assuming this bug is caused by some weird developer interaction.
- **beta and prod databases have not been created on the Firebase console yet**
    - additionally, we should double-check that **read and write access to these databases are set to true** for testing, otherwise, an error will be thrown and cause a crash  
  - Unit and integration tests need to be written so we can automatically verify that forms are working as intended, data is being uploading to Firebase correctly, and that we can then read data from an API request to the database. Issue #199 is specifically for this issue
  - I have not extensively tested the address autocomplete functionality, and there may be edge cases or issues related to network/latency I'm not aware of.
  - **all state of the form is uploaded to every resource type**. this means that each resource type has information that's not relevant to that resource type (e.g. `foraging_food_type` field existing on a `water` entry). This was for convenience, as client-side code is not affected by these extra fields existing on the resource entry, **but I recognize this might lead to latency issues as there's more data per entry on each resource**
      - one way to address this is writing additional logic to check what resource type we are submitting, and constructing the appropriate state object for that resource type, and submitting that (as opposed to just submitting the whole form state). This can be found in the `onSubmit` function in `AddResourceModal.js`


gcardonag (approved):

affabillyty (approved):
Wahoo! 🚀

PR #203, opened 2021-10-12
Misc: Create `schema.graphql` 

nwyin:
# Pull Request

## What

Create a `graphql.schema` file that contains the fields and types for the crowdsourcing form

## Why

A way for developers/non-developers to view the schema without having to request access to the private Google Sheet. Has additional benefits that:
- schema is versioned control, so we can see how the data we're collecting has changed over time
- could serve as the one source of truth for our data moving forward, which might hopefully reduce developer friction moving forward

### DIFF
--- a/api/schema.graphql
+++ b/api/schema.graphql
@@ -1,0 +1,100 @@
+# Types = Sheet/Table in Excel
+type Example {
+  id: Int # fields are columns
+}
+
+enum Access {
+  PUBLIC
+  PRIVATE_SHARED
+  PRIVATE
+  RESTRICTED
+}
+
+enum YesNoUnsure {
+  YES
+  NO
+  UNSURE
+}
+
+interface SharedFields {
+  description: String
+  phlaskStatement: String
+  normsAndRules: String
+  website: String
+  additionalInfo: String
+}
+
+enum WaterTapTypes {
+    DRINKING_FOUNTAIN
+    BOTTLE_FILLER
+    SINK
+    SODA_FOUNTAIN
+    WATER_DISPENSER
+    WATER_COOLER
+    OTHER
+}
+
+type WaterTap implements SharedFields {
+  access: Access!
+  name: String!
+  address: String!
+  hours: [String] # array of hours of availbility -- each entry is day of week, [Monday Hours, Tuesday Hours, Wednesday...]
+  service: SELF_SERVE | ASK_PROPRIETOR | UNSURE
+  type: WaterTapTypes
+  filtration: YesNoUnsure
+  accessible: YesNoUnsure
+  sparkling: YesNoUnsure
+  vesselNeeded: YesNoUnsure
+  idRequired: YesNoUnsure
+}
+
+enum FoodAccess  {
+    GOVERNMENT
+    BUSINESS
+    NONPROFIT
+    RELIGIOUS
+    GRASSROOTS
+    OTHER
+}
+
+enum FoodConsumptionTypes {
+    ONSITE
+    TAKEAWAY
+    DELIVERY
+}
+
+type Food implements SharedFields {
+  access: FoodAccess!
+  name: String!
+  address: String!
+  hours: [String] # array of hours of availbility -- each entry is day of week, [Monday Hours, Tuesday Hours, Wednesday...]
+  consumption: FoodConsumptionTypes
+  foodType: [PERISHABLE | NONPERSIHABLE | PREPARED | OTHER]  # idk if this works
+  accessible: YesNoUnsure
+  idRequired: YesNoUnsure
+  minorsOnly: YesNoUnsure
+}
+
+enum ForagingFoodTypes {
+    NUT
+    FRUIT
+    LEAVES
+    BARK
+    FLOWERS
+    MUSHROOM
+    ROOT
+    OTHER
+}
+
+type Foraging implements SharedFields {
+  access: Access!
+  name: String!
+  address: String!
+  hours: [String] # array of hours of availbility -- each entry is day of week, [Monday Hours, Tuesday Hours, Wednesday...]
+  genus: String
+  species: String
+  features: [SWEET | SAVORY | MEDICINAL] # idk if this works
+  type: ForagingFoodTypes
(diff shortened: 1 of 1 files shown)


nwyin:
Please confirm with @affabillyty if this seems like a reasonable `graphql.schema` to merge -- can always make commits to change or future PRs to change it as it evolves/needs changes.

nwyin:
Closing since no one has looked at this in a while. Can return to this in the future if necessary.

api/schema.graphql
```diff
@@ -0,0 +1,100 @@
+# Types = Sheet/Table in Excel
+type Example {
+  id: Int # fields are columns
+}
+
+enum Access {
```
gcardonag, line 11: For consistency with other resource types, we should change this to `WaterAccess`. We should also add the following values to align with what we have in our data today:
```
WM
TRASH_ACADEMY
```

api/schema.graphql
```diff
@@ -0,0 +1,100 @@
...
+  PUBLIC
+  PRIVATE_SHARED
+  PRIVATE
+  RESTRICTED
+}
+
+enum YesNoUnsure {
+  YES
+  NO
+  UNSURE
```
gcardonag: I'm thinking we should add a comment to highlight that `UNSURE` should be the default value when using this enum. Maybe something as simple as `UNSURE # Default Value`. Open to other ideas here. I read how GraphQL defines default argument values and find that the way its expressed may just end up confusing folks who may think we always set it to `UNSURE`.

api/schema.graphql
```diff
@@ -0,0 +1,100 @@
...
+  name: String!
+  address: String!
+  hours: [String] # array of hours of availbility -- each entry is day of week, [Monday Hours, Tuesday Hours, Wednesday...]
+  service: SELF_SERVE | ASK_PROPRIETOR | UNSURE
+  type: WaterTapTypes
+  filtration: YesNoUnsure
+  accessible: YesNoUnsure
+  sparkling: YesNoUnsure
+  vesselNeeded: YesNoUnsure
+  idRequired: YesNoUnsure
```
gcardonag, line 55: We can remove this field for now since we've never had that requirement for a water location. It makes more sense for food banks.

api/schema.graphql
```diff
@@ -0,0 +1,100 @@
...
+    OTHER
+}
+
+type WaterTap implements SharedFields {
+  access: Access!
+  name: String!
+  address: String!
+  hours: [String] # array of hours of availbility -- each entry is day of week, [Monday Hours, Tuesday Hours, Wednesday...]
+  service: SELF_SERVE | ASK_PROPRIETOR | UNSURE
+  type: WaterTapTypes
```
gcardonag: This should be changed to `[WaterTapTypes]` as a location can have more than one type of water dispenser, judging by our data. For example, I found `Drinking fountain and water dispenser` and `Bottle filler and fountain` examples.

api/schema.graphql
```diff
@@ -0,0 +1,100 @@
...
+    WATER_DISPENSER
+    WATER_COOLER
+    OTHER
+}
+
+type WaterTap implements SharedFields {
+  access: Access!
+  name: String!
+  address: String!
+  hours: [String] # array of hours of availbility -- each entry is day of week, [Monday Hours, Tuesday Hours, Wednesday...]
```
gcardonag: I would recommend we represent hours as something like `[AvailabilityBlock]`, where `AvailabilityBlock` is:
```
type AvailabilityBlock {
  open: String!
  close: String!
}
```

Also we should update the comment to start with `Sunday Hours`, since we treat the 0-index of the list as Sunday: https://github.com/phlask/phlask-map/blob/develop/src/helpers/hours.js#L90

api/schema.graphql
```diff
@@ -0,0 +1,100 @@
...
+    ONSITE
+    TAKEAWAY
+    DELIVERY
+}
+
+type Food implements SharedFields {
+  access: FoodAccess!
+  name: String!
+  address: String!
+  hours: [String] # array of hours of availbility -- each entry is day of week, [Monday Hours, Tuesday Hours, Wednesday...]
```
gcardonag: Same comment here as the previous comment about how we should structure `hours` values for Water data points.

api/schema.graphql
```diff
@@ -0,0 +1,100 @@
...
+enum FoodAccess  {
+    GOVERNMENT
+    BUSINESS
+    NONPROFIT
+    RELIGIOUS
+    GRASSROOTS
+    OTHER
+}
+
+enum FoodConsumptionTypes {
```
gcardonag, line 67: We'll need an `UNSURE` value here to use as a default for this field since we may not have that data yet for some Food data points.

gcardonag (commented):
Thanks again for the work on this! I dug through our existing data and the values for our fields to come up with some suggestions on some adjustments before we merge this. In addition, we'll need to make a few changes to our data to align to this schema, which I think is worth doing (I can work on that):

For Water
- filtration
    - Set blanks to UNSURE
- handicap
    - Rename to accessible
    - Set blanks to UNSURE
- sparkling
    - Set blanks to UNSURE
- vessel
    - Rename to vesselNeeded
    - Set blanks to UNSURE
- tap_type
    - Rename to "type"
    - Set blanks to OTHER
- service
    - Set blanks to UNSURE

For Food
- access
    - Need to revisit current types ('Charter School', 'Food Site', 'PHA Community Center', 'School') to align them with our new types (GOVERNMENT, BUSINESS, NONPROFIT, RELIGIOUS, GRASSROOTS, OTHER)
- organization
    - Rename to "name"
- zip_code
    - Merge values into "address" field
- city
    - Merge values into "address" field
- consumption
    - Set to UNSURE by default
- foodType
    - Set to OTHER by default
    - We'll need a story to go through the data and set this value since we don't have it yet. The data seems to be in the description for some locations.
- accessible
    - Set to UNSURE by default
- id_required
    - Change to idRequired
    - Set to UNSURE by default
- kid_only
    - Change to minorsOnly
    - Set to UNSURE by default

gcardonag (approved):

PR #243, opened 2022-03-10
Adding ThemeProvider for MUI Components

edlopez000:
Introduces a custom [ThemeProvider](https://mui.com/customization/theming/#theme-provider) in theme.js that is aligned with our [Design System](https://www.figma.com/file/VGGqwl3Eq3GQIAO9I6LoNW/phlask?node-id=2786%3A10906). 

For the Design System font scale, the theme.js is taking advantage of the new [responsiveFontSizes](https://mui.com/customization/theming/#responsivefontsizes-theme-options-theme) function which utilizes breakpoints to handle font scaling. Currently set to the default breakpoints. The values will need to be adjusted, but the breakpoints are fully functioning as shown in the GIF below:

ReponsiveFonts

This PR also adds [CSSBaseline](https://mui.com/components/css-baseline/) to the main app to normalize the CSS across different browsers.


nwyin:
@edlopez000 Thanks for this work! Would you be able to include any screenshots/GIFs/short videos demoing the feature?@gcardonag will sometimes pull down code & test locally, but it's useful to have some visuals in the PR for other folks who aren't reviewers to be able to see what the changes do.

Also helps out so reviewers can have an easier time identifying what's the expected behavior should be.

src/App.js
```diff
@@ -33,6 +35,7 @@ function App(props) {
   ReactGA.pageview(window.location.pathname + window.location.search);
   return (
     <ThemeProvider theme={theme}>
+      <CssBaseline />
```
nwyin, line 38: Thanks for adding this -- it's a standard practice to include this on sites/web apps nowadays but we've neglected this (until now)

src/theme.js
```diff
@@ -1,27 +1,110 @@
...
+      fontSize: ".579rem"
+    },
+    button2: {
+      fontWeight: 600,
+      fontSize: "1.2rem"
+    }
+  }
+});
+
+theme = responsiveFontSizes(theme);
```
nwyin, line 108: Could you annotate this line describing what `responsiveFontSizes` does? I did read the doc pages you sent, but this is for the benefit of @gcardonag when he takes a look at it, and also because having in-line annotations for these sorts of things is useful for when you need to look back on PR's to figure out WTH you were changing and why.

nwyin (changes requested):
Changes look good! Just had a few requests about some annotations/documentation that would be great to add.

However, reach out to @gcardonag first. If he feels like the extra annotations + documentation isn't worth it, then feel free free to re-request my review, point me to his comment, and I'll approve the PR too.

Additionally, make sure builds are passing (https://github.com/phlask/phlask-map/runs/5500019398). Won't be able to merge in otherwise. Let's take a look at this together during hacknight.

edlopez000:
> @edlopez000 Thanks for this work! Would you be able to include any screenshots/GIFs/short videos demoing the feature?@gcardonag will sometimes pull down code & test locally, but it's useful to have some visuals in the PR for other folks who aren't reviewers to be able to see what the changes do.
> 
> Also helps out so reviewers can have an easier time identifying what's the expected behavior should be.

@tnguyen21 I added GIF showing how the breakpoints function.

src/theme.js
edlopez000, line 108: responsiveFontSizes() is Material UI's way of implementing responsive font scaling. Based on the width of the screen, the font will increase or decrease. We can set both when (breakpoints) the font will increase/decrease and by how much (nested `fontSize` with the rem value) as well.

edlopez000: I went back and added an explanation of what responsiveFontSizes, please let me know if that helps explain the function.

nwyin: @edlopez000 LGTM! I'll go ahead and approve your PR. It's good manners to let the PR author click "Merge" themselves, so I'll leave that to you.

(Although, if it's an urgent fix that looks correct, then often someone else can/will merge it in to get the fix in ASAP)

nwyin (approved):
LGTM

PR #265, opened 2022-06-28
minor refactor of getTaps and getFoodOrgs

RyanHarlow:
# Pull Request

## Change Summary
In action.js in the functions getTaps and  getFoodOrgs instead of calling snapshot.val() multiple times inside of loop, refactored to call snapshot.val() only once and store the returned value as a variable, which can then be used inside the loop

## Change Reason
This change improves initial page load time by several seconds. It seems snapshot.val() is a relatively time-consuming function and it was being called a few hundred times inside of the getTaps() function. This refactor instead stores the value returned from firebase by snapshot.val() to a variable which can be accessed much more efficiently improving page load time

## Verification [Optional]

Lighthouse result before refactor:
lighthouse-original

Lighthouse result after refactor:
lighthouse-after-refactor

Related Issue: #<Github Issue Number (if one exists)>

### DIFF
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -65,19 +65,20 @@ export const getTaps = () => dispatch => {
   const app = initializeApp(waterConfig, 'water');
   const database = getDatabase(app);
   return onValue(ref(database, "/"), (snapshot) => {
+      const snapshotVal = snapshot.val();
       var allTaps = [];
       var item;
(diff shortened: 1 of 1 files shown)


gcardonag (approved):
Woah, that's a huge time save with a minimal change! Thanks for this 🎉

PR #272, opened 2022-08-09
build(format): prettier configuration

RNR1:
# Pull Request

## Change Summary

I added a specific prettier configuration to ensure that any developer in the dev circle will follow those formatting settings.
I also run the `npm run format` script to ensure all the source code follows the new guidelines.

## Change Reason

Lacking a set of code formatting configurations, we encounter a situation where any accidental visit of a file will end up in our code editor trying to "format on save" using the editor format settings.
It will cause changes in a file we didn't work on, and end up in inconsistent code formatting  + confusion when reviewing new pull requests, as it's hard to figure out what changes are format-only changes and what changes are related to the actual issue.


gcardonag (approved):
The site built properly during the test. Approving! 👍

For reference for others, the reason the test failed is because the code came from a fork, which doesn't allow access to our AWS credentials, failing the S3 upload step.

Issue #277, opened 2022-08-24
Reduce our use of class-based components


gcardonag:
**Is your feature request related to a problem? Please describe.**
As a PHLASK developer, I would like to see a more consistent set of react code as I try to make changes. In order to support this, I would like to have more functional components, as we are starting to see this more as the preferred option for newer developers and in code samples in the wild.

**Describe the solution you'd like**
More functional code samples, but having the site still functioning as expected.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

- [x] MapMarkersFood.js
- [x] MapMarkers.js
- [x] AddResourceModal.js
- [x] MapPage.js
- [x] ReactGoogleMaps.js
- [x] SearchBar.js
- [x] SelectedTap.js

gcardonag:
![](https://github.trello.services/images/mini-trello-icon.png) [Migrating class components to functional components in react](https://trello.com/c/tkmNwAKQ/4-migrating-class-components-to-functional-components-in-react)

PR #278, opened 2022-08-24
refactor(map-markers): migrate map marker food to fc

RNR1:
# Pull Request

## Change Summary

- Migrated the MapMarkersFood.js component from class to function component
- Used React.memo in order to prevent re-renders of the center map marker (previously used with the "shouldComponentUpdate" lifecycle method)
- Used useEffect in order to conditionally fetch food orgs if not available (previously used the "UNSAFE_componentWillMount" lifecycle method
- Removed unused methods

## Change Reason

According to issue #277, we are reducing our usage of class based components, which are part of an old React API and harder to maintain.

## Verification [Optional]
Provided a gif in the comments

Related Issue: #277

### DIFF
--- a/src/components/MapMarkers/MapMarkersFood.js
+++ b/src/components/MapMarkers/MapMarkersFood.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React from 'react';
 import { Marker } from 'google-maps-react';
 import { connect } from 'react-redux';
 import {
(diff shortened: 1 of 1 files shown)


RNR1:
phlask-map-food-marker

gcardonag (approved):
Tested it out on the test site and it works like a charm, this is awesome! 🎉

PR #279, opened 2022-08-31
refactor(map-markers-water): migrate to fc

RNR1:
# Pull Request

## Change Summary

- Migrated the MapMarkers.js component from class to function component
- Used React.memo in order to prevent re-renders of the center map marker (previously used with the "shouldComponentUpdate" lifecycle method)
- Used useEffect in order to conditionally fetch food orgs if not available (previously used the "UNSAFE_componentWillMount" lifecycle method
- Removed unused methods

## Change Reason

According to issue https://github.com/phlask/phlask-map/issues/277, we are reducing our usage of class based components, which are part of an old React API and harder to maintain.

## Verification [Optional]

phlask-markers

Related Issue: #277

### DIFF
--- a/src/components/MapMarkers/MapMarkers.js
+++ b/src/components/MapMarkers/MapMarkers.js
@@ -10,148 +10,34 @@ import {
 } from '../../actions/actions';
 import makeGetVisibleTaps from '../../selectors/tapSelectors';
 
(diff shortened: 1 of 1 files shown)


gcardonag (approved):
Thanks again for making our code more functional 😄

Issue #304, opened 2023-02-26
Create V.2 FilterDrawer (Desktop)


gcardonag:
**Is your feature request related to a problem? Please describe.**
As a part of the effort to deploy PHLASK 2.0, we would like the FilterDrawer component in the src/components/FilterDrawer folder to match the design in the Desktop design shared here: https://www.figma.com/proto/VGGqwl3Eq3GQIAO9I6LoNW/PHLASK?page-id=4510%3A29556&type=design&node-id=8105-140052&viewport=889%2C-851%2C0.09&scaling=scale-down&starting-point-node-id=8105%3A139899&show-proto-sidebar=1&mode=design 

**Describe the solution you'd like**
A user accessing PHLASK in a Desktop view experiences a design matching the one proposed for PHLASK 2.0. This issue is focused on the Filter Drawer shown in the bottom-left of the linked Figma design.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

jordanforbes:
I've actually been working on this. Right now we don't have the full list of suggested filters implemented in the redux store so I only have the three that work on the filter at the moment. It does work, however.

gcardonag:
For anyone working on this, I updated the drawer visual target to a newer version with a prototype that covers filters for each resource type.

PR #312, opened 2023-03-28
Filter dev

jordanforbes:
# Pull Request

## Change Summary

The filter drawer now has buttons. When the buttons are pressed the redux state is changed, altering which locations are shown on the map. 

## Change Reason

It's currently at a working state although only 3 of the suggested filters are currently implemented in the redux store itself. 

Related Issue: #304


gcardonag:
Hey @jordanforbes ! I know we've circled around this Pull Request, but is this one ready for review? I want to make sure to go through our older Pull Requests first if they are ready and try and keep better track of the work we've been up to via comments as I sometimes need to pull away from weekly reviews and lose track of history. Sorry for the late follow-up here 🙇

PR #316, opened 2023-05-17
Joe/v2 toolbar

JosephAni:
# Pull Request


gcardonag:
Skipping the test for this one as it will normally fail due to forks not having access to get our AWS credentials

gcardonag (approved):
Thanks for the fancy new toolbar 🥳

JosephAni:
🤩

Issue #302, opened 2023-02-26
Create V.2 SearchBar (Desktop)


gcardonag:
**Is your feature request related to a problem? Please describe.**
As a part of the effort to deploy PHLASK 2.0, we would like the SearchBar component in the src/components/SearchBar folder to match the design in the Desktop design shared here: https://www.figma.com/file/VGGqwl3Eq3GQIAO9I6LoNW/PHLASK?node-id=6451%3A77014&t=xE1rkr657wCNubGL-4

**Describe the solution you'd like**
A user accessing PHLASK in a Desktop view experiences a design matching the one proposed for PHLASK 2.0. This issue is focused on the Searchbar shown in the bottom-left of the linked Figma design.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

Erw1nn:
I asked if I could work on this at Hack Night and they gave me the green light, so I assigned it to myself 🙂

gcardonag:
Closing this one as complete 🎉

PR #319, opened 2023-05-24
Task 302: Create V.2 SearchBar (Desktop)

Erw1nn:
# Pull Request

## Change Summary

Moved Search bar to the bottom of the page, on top of the toolbar. Also moved information button slightly to the left. Since it's in the same container as the search bar, it would have overlapped with the zoom buttons if it wasn't moved since I moved the container down

## Change Reason

Implementing task 302: Create V.2 SearchBar (Desktop)

## Verification [Optional]

New Look:

image

Related Issue: #302

### DIFF
--- a/src/components/ReactGoogleMaps/ReactGoogleMaps.js
+++ b/src/components/ReactGoogleMaps/ReactGoogleMaps.js
@@ -2,6 +2,7 @@ import { Map, GoogleApiWrapper, Marker } from 'google-maps-react';
 import React, { Component } from 'react';
 import ReactTouchEvents from 'react-touch-events';
 import SearchBar from '../SearchBar/SearchBar';
+import TutorialModal from '../TutorialModal/TutorialModal';
 import styles from './ReactGoogleMaps.module.scss';
 import { connect } from 'react-redux';
 import SelectedTap from '../SelectedTap/SelectedTap';
@@ -23,6 +24,7 @@ import MapMarkersFood from '../MapMarkers/MapMarkersFood';
 import { isMobile } from 'react-device-detect';
 import Toolbar from '../Toolbar/Toolbar';
 import MapMarkersMapper from '../MapMarkers/MapMarkersMapper';
+import Stack from '@mui/material/Stack';
 
 // // Actual Magic: https://stackoverflow.com/a/41337005
 // // Distance calculates the distance between two lat/lon pairs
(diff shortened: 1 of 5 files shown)


src/components/ReactGoogleMaps/ReactGoogleMaps.module.scss
```diff
@@ -1,6 +1,6 @@
 .searchBarContainer {
   position: relative;
-  top: 57px;
+  top: 608px;
```
IAmConnorWilson: I'm not 100%  since I don't have this running locally but you might want to do this from the bottom instead of top or you might get inconsistent results based on the screen sizes

src/components/SearchBar/SearchBar.js
```diff
@@ -86,13 +89,18 @@ class SearchBar extends React.Component {
...
-                  <input
+                  <Input
                     {...getInputProps({
-                      placeholder: 'Search For Taps Near...'
+                      placeholder: 'Search For Resources Near...'
                     })}
                     className={`${styles.searchInput} form-control`}
                     type="search"
                     ref={this.state.refSearchBar}
+                    startAdornment={
```
IAmConnorWilson, line 98: nice!

src/components/ReactGoogleMaps/ReactGoogleMaps.module.scss
Erw1nn: Thanks for the feedback. My most recent commit does it from the bottom now

src/components/SearchBar/SearchBar.js
Erw1nn: Thanks!

IAmConnorWilson (approved):
looks Great!

src/components/ReactGoogleMaps/ReactGoogleMaps.js
```diff
@@ -342,10 +344,15 @@ export class ReactGoogleMaps extends Component {
           </div>
         </ReactTouchEvents>
         <div className={styles.searchBarContainer}>
```
gcardonag: I like the idea of using the `Stack` component to group elements horizontally! Could we try to apply it to vertically stack things on the bottom-left of the Desktop view? I'm thinking this may be a way to simplify things from a CSS perspective.

Erw1nn: I added a Stack to group the items in the bottom-left of the Desktop view. Let me know how it looks

PR #359, opened 2023-12-13
Changing resource now also changes filter

astoppop:
# Pull Request

## Change Summary

Changing resource now also changes the filter. Buttons on the filter do work and you can apply them. However, they don't actually do anything yet as showing non-water resources hasn't been fully implemented.

## Change Reason

Prior to this change, selecting any non-water resource wouldn't change the filter and it would still show the filter tags as though your selected resource was water

## Verification [Optional]

Before: 
image

After:
image

### DIFF
--- a/src/components/Filter/Filter.js
+++ b/src/components/Filter/Filter.js
@@ -12,12 +12,12 @@ const FilterTags = props => (
         key={key}
         className={
           styles.filterTag +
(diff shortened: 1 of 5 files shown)


gcardonag (approved):
Thank you for this change @astoppop ! The updated filters look great 🙇

Issue #357, opened 2023-12-05
Two markers for the Indy hall address on the search bar


mandyshri101:
**Describe the bug**
Two markers populating after entering Indy hall address on the search bar. Picture attached for the reference

**To Reproduce**
1. Go to phlask.me
2. Click on search Indy hall and select it and press enter
3. Scroll down to google maps area
4. You would find the two makers for the same address.

**Expected behavior**
Indy hall with "Indy Hall Clubhouse at 709 N 2nd St, North 2nd Street, Philadelphia, PA, USA" entry should have only one marker of the google marker as show in the below expected screenshot marker

**bug Screenshots**
<img width="1440" alt="image" src="https://github.com/phlask/phlask-map/assets/143658338/cff671b9-dc9e-4d8e-975d-7ee386087c27">
**Expected screenshot**
<img width="1421" alt="image" src="https://github.com/phlask/phlask-map/assets/143658338/31ac80b0-3fcd-41bf-ba55-59ac5db94c87">

**Desktop (please complete the following information):**
 - OS: Mac OS
 - Browser Chrome
 - Version 120.0.6099.62

**Smartphone (please complete the following information):**

**Additional context**

vontell:
I'm happy to take on this task! Seems like an additional marker is being placed on the pin for Philadelphia as a whole. Not sure how to assign myself to this, but would like to take it on!

vontell:
@mandyshri101 Dug into this issue more, and I found that two markers aren't actually appearing for a search - what's happening is that when the map and your location loads, there is a default pin being placed at the center of your map (I noticed this because with my location services off, it places a pin on Philadelphia, but when turned on (and I'm traveling) it placed a pin at the center of my current location).

I looked at the history of this code, and it's been there for a long time (the last change to that block of code was a year ago during a refactor) so I can't determine if the change is intentional or not. I personally think it doesn't make much sense to show their current location as a pin - instead we should enable a blue dot indicator or some other smaller icon.

Opened a PR here: https://github.com/phlask/phlask-map/pull/360

Here is a screenshot with the change:
<img width="1728" alt="Screenshot 2023-12-16 at 12 31 27 PM" src="https://github.com/phlask/phlask-map/assets/8053203/f4aa47d4-2a2f-4ea0-9bbf-71935d4b7fb8">

PR #360, opened 2023-12-16
[Issue 357] Remove extraneous marker from map

vontell:
# Pull Request

Fix for #357 (see issue for full context)

## Change Summary

Dug into this issue more, and I found that two markers aren't actually appearing for a search - what's happening is that when the map and your location loads, there is a default pin being placed at the center of your map.

I looked at the history of the code that caused this, and it's been there for a long time (the last change to that block of code was a year ago during a refactor) so I can't determine if the change is intentional or not. I personally think it doesn't make much sense to show their current location as a pin - instead we might want to enable a blue dot indicator or some other smaller icon.

## Change Reason

Change fixes a bug tracked by issue #357 

## Verification [Optional]

Here is a screenshot with the change:
Screenshot 2023-12-16 at 12 31 27 PM

Related Issue: #357

### DIFF
--- a/src/components/MapMarkers/MapMarkers.js
+++ b/src/components/MapMarkers/MapMarkers.js
@@ -31,13 +31,6 @@ const MapMarkers = ({
 
   return (
     <>
-      <Marker
-        map={map}
-        google={google}
-        key="current_pos"
-        name="Current Pos"
-        position={mapCenter}
-      />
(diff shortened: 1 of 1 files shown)


gcardonag (approved):
Thank you for cleaning this up! We'll likely need a better way to represent the user's location down the line.

PR #407, opened 2024-03-27
Complete Acknowledgements Page with Live Data from Firebase

### LINKED ISSUE #355: [DRAFT] Complete Acknowledgements Page with Live Data from Firebase
We would like to have an Acknowledgements section available in our Contribute page, such that whenever the page is loaded, a dynamic list of contributors is generated via a reference to our phlask-contributors Firebase database. This should be presented as (relative to firebase fields):
```
id
First_Name Last_Name
Circle Names (the ones set to true)
"Convener" (only displayed if the Convener property is true)
Status
```
Only the users with a status set to "Active" or "Engaged" in Firebase should be displayed.

Pull Request #354 added the Acknowledgements page, a route to the page, and firebase configuration on the frontend for issue #294.  This new ticket seeks to take the feature further by pulling data from firebase and presenting the information on the acknowledgements page.  

The style of the pages is TBD, and a link to the figma design will be posted here when available.  

n/a 

**Additional context**
n/a

RNR1:
# Pull Request

## Change Summary

Added the acknowledgment section on the "Join The Team" page

## Change Reason

Product request

## Verification [Optional]
image

Related Issue: #355


src/components/ContributorsList/ContributorsList.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import List from '@mui/material/List';
```
ravimakhijapi, line 2: Small nit/question - it looks like these imports can be shortened to:
`import { List, Grid, Typography } from '@mui/material';`

It seems the material-ui docs tend to use individual import paths (like you have currently) in their examples. On the other hand, importing from `'@mui/material'` is concise. 

Any preference/thoughts?

src/components/AddResourceModal/AddBathroom.jsx
```diff
@@ -113,7 +113,7 @@ function AddBathroom({
     }
   ];
 
-  const userLocation = useSelector(state => state.userLocation);
+  const userLocation = useSelector(state => state.filterMarkers.userLocation);
```
ravimakhijapi, line 116: I was thinking it might make sense split `filterMarkers.js` into different files and rename appropriately to reflect the contents better. In particular, the action `SET_TOOLBAR_MODAL` doesn't seem to fit into the "filterMarkers" category, I think. Any thoughts? 

Since this PR will require changing the approach to referencing the `filterMarkers` reducer, it seemed like a good time to bring this up, though it may make sense to address in a separate follow-up issue if we decide to do that.

RNR1: It makes sense, but it will be a significant shift for this, as we need to map all state fields into new slices, and touching this part of the code can be very fragile. 
I think it's worth having a refactor ticket that will be dedicated to this body of work

src/components/ContributorsList/ContributorsList.js
RNR1: That's a good point! The reason for that is documented [here](https://mui.com/material-ui/guides/minimizing-bundle-size/)!

TL;DR - default direct imports are faster in the development environment and provide a smaller bundle size - though the latter will not matter in our case when we use a bundler that supports tree-shaking

gcardonag (approved):
LGTM!

src/components/AddResourceModal/AddBathroom.jsx
gcardonag: I drafted an issue to follow up on this thread here: https://github.com/phlask/phlask-map/issues/432

Issue #392, opened 2024-02-21
[V2] Populate lat/lng properly before form submission


vontell:
This task is all about adding **latitude and longitude** information to the resource submission page.

First, take a look at [AddResourceModalV2.jsx - line 231](https://github.com/phlask/phlask-map/blob/develop/src/components/AddResourceModal/AddResourceModalV2.jsx#L231-L302). You'll see here that we take the information from the add resource forms, and then submit it to our firebase DB.

You'll also notice, however, that the latitude and longitude is not included here. Our goal for this task is to add the user's latitude and longitude for that location.

Inside of the [AddResourceModalV2.jsx](https://github.com/phlask/phlask-map/blob/develop/src/components/AddResourceModal/AddResourceModalV2.jsx), we can access the user's location using:

```
const userLocation = useSelector(state => state.userLocation);
```

You can search for other instances of `userLocation` in the code to see what its format looks like. What you can do is then add a `latitude` and `longitude` field to the JSON being submitted in that on submit button.

This is the first step in our task of supporting the addition of this information - in another follow up issue and PR, we will also need to make sure that the user has location enabled.

To verify your work, you should console.log the JSON before it gets sent to firebase, and you should be able to view the networking profiler in chrome and see if a request was sent with the lat/long.

I am happy to hop into a call and help debug and review! I assume that will be needed since this will be your first task.

freelyPheobe:
Currently going through testing.

freelyPheobe:
-Only allowed to ask user location once.
- Addresses can be associated with Parks and the like.  Thus utilizing a precise lat/long gives a more accurate position.
-Also- we have to plan for users that add locations other than their current location (ex. water fountains in a mall).

-Version 2. XX coming to a Phlask near you.

tomporvaz:
@vontell and @Tom-from-6520 I wondering how this ticket it going.  Please comment an update on the status of this ticket.  Is Tom still working on this?

vkavith:
I will take this ticket

vkavith:
Is Tom-from-6520 already working on this ?

tomporvaz:
@vkavith Lets see if we get a response from @Tom-from-6520 or @vontell by next Tuesday 4/23. Vontell could probably give us an update at the meeting.

Tom-from-6520:
Hi, I apologize for holding on to this ticket. I'm not working on it currently.

vontell:
[...]

PR #437, opened 2024-05-29
Geocode Address in Form

### LINKED ISSUE #426: Give forms the ability to set either the address location or current location

In a previous ticket (#392), we added the existing latitude and longitude for a resource into the Firebase data. However, we so far only implemented the option of "use my existing address", not the address of an address typed in. This ticket is to make sure that the user has both options.

The logic should be as follows:
- If the user provides an address (by searching or typing), we should use that latitude and longitude from the google places API.
- If they don't type out an address and click the "use my location" button, it should use that exact lat and long, **and** also populate the new address at that location into the Street address field.

kdmonroe:
# Pull Request

## Change Summary

- Added getLatLng from [react-places-autocomplete](https://www.npmjs.com/package/react-places-autocomplete#geocode-by-address) 
- Updated `textFieldChangeHandler` to geocode the address (using the above library). Uses a debounce delay. 
- Added lat/lon fields to `initialState` (as null) and `newData` json (updated).

## Change Reason

- Related issue #392 addresses handling the user location and reverse geocoding. But that was the default lat/lng being passed as the resource location as well. With this PR, the user can specify an address (and its geocoded lat/lon values are passed with new data inserts) 

## Verification [Optional]

- [Video Explanation](https://drive.google.com/file/d/1zWj-tqIm6RnMqt7l6lW6KZB4domphxP6/view?usp=drive_link)

Related Issue: #392

### DIFF
--- a/src/components/AddResourceModal/AddResourceModalV2.jsx
+++ b/src/components/AddResourceModal/AddResourceModalV2.jsx
@@ -14,7 +14,7 @@ import AddBathroom from './AddBathroom/AddBathroom';
 import AddForaging from './AddForaging/AddForaging';
 import AddWaterTap from './AddWaterTap/AddWaterTap';
 import ModalWrapper from './ModalWrapper';
-import { getDatabase, ref, set, onValue } from 'firebase/database';102
+import { getDatabase, ref, set, onValue } from 'firebase/database';
 import { geocodeByAddress, getLatLng } from 'react-places-autocomplete';
 
 export default function AddResourceModalV2(props) {
--- a/src/components/AddResourceModal/AddResourceModalV2.jsx
+++ b/src/components/AddResourceModal/AddResourceModalV2.jsx
@@ -6,6 +6,7 @@ import {
   TOOLBAR_MODAL_CONTRIBUTE
 } from '../../actions/actions';
 
+import { debounce } from '../../utils/debounce';
 import ChooseResource from './ChooseResource';
 import ShareSocials from './ShareSocials';
 import AddFood from './AddFood/AddFood';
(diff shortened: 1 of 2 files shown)


src/components/AddResourceModal/AddResourceModalV2.jsx
```diff
@@ -92,25 +95,70 @@ export default function AddResourceModalV2(props) {
...
-      // the address textbox is set with react-hook-form setValue() when user clicks "use my location instead",
-      //  which doesn't fire an event. When that happens, a string is passed instead of an event object, so this
-      // conditional expression handles that
-
-      if (e.target) {
-        return {
-          ...prevValues,
-          [e.target.name]: e.target.value
-        };
+  const debounce = (func, delay) => {
```
RNR1: Can we move this to a new file `src/utils/debounce.js`?

src/components/AddResourceModal/AddResourceModalV2.jsx
```diff
@@ -92,25 +95,70 @@ export default function AddResourceModalV2(props) {
...
+const debouncedGeocode = debounce(address => {
+  geocodeByAddress(address)
+    .then(results => {
+      if (results.length === 0) {
+        throw new Error('ZERO_RESULTS');
+      }
+      return getLatLng(results[0]);
+    })
+    .then(({ lat, lng }) => {
+      console.log('Successfully got latitude and longitude', { lat, lng });
```
RNR1: Do you think we can remove this console statement? It's useful for local development but doesn't provide value afterward.

RNR1 (approved):
LGTM with a few comments

src/components/AddResourceModal/AddResourceModalV2.jsx
kdmonroe: Hi @RNR1 - I moved the debounce function to `src/utils/debounce.js` and updated the import statement accordingly. Also, I removed the console.log statements as requested. Please review the changes. Thanks!

PR #438, opened 2024-06-04
[Issue 419] address firebase concurrency concerns

patrickcbrady:
# Pull Request

## Change Summary

Use automatic id generation in firebase to ensure that identifiers from multiple instances writing simultaneously will not overlap.

## Change Reason

Individual instances currently manually increment an identifier and send it directly, creating a race condition in the case where two instances write new entities close in time.

## Verification [Optional]

Adding entries to existing DB
Screenshot 2024-06-05 at 6 35 04 AM

Adding entry to empty DB
Screenshot 2024-06-05 at 6 42 34 AM


vontell (approved):
Nice! Glad it was a simple change, thanks for looking into this.

PR #428, opened 2024-05-01
Added Latitude and Longitude Use my Location setting but have to enable address-based lat/lng

vkavith:
# Pull Request

## Change Summary

Summary of changes made. This can be a paragraph, list of bullet points, etc. Aim to be brief and concise.

## Change Reason

Summary of why changes are being made. Link to other documents, Slack conversations, etc are good to post here for explaining the reason for particular changes.

## Verification [Optional]

A .gif, video, or screenshots of the feature are extremely helpful in verifying that code changes are working correctly. Feel free to include them in the PR to help the reviewer.

We recommend using [OBS](https://obsproject.com/) to record videos, as it is open source, free, and available on all major operating systems.

Related Issue: #<Github Issue Number (if one exists)>

Issue #228, opened 2021-12-12
Create V.2 SearchBar Mobile 


dibleauzah:
Is your feature request related to a problem? Please describe.
As a user, 
I want to search for resources, landmarks, or addresses 
in order to find the best resource near me/around a certain location. 

Describe the solution you'd like
- A search bar with auto-fill addresses.
- Searchable by landmarks (nice to have)
- Searchable by Tap Name
- Searchable by resource type (water, food, toilets, etc. with synonyms) (nice to have)

Describe alternatives you've considered
XX

Additional context
[Figma Page](https://www.figma.com/file/VGGqwl3Eq3GQIAO9I6LoNW/PHLASK?type=design&node-id=3427-32115&mode=design&t=AXQfJKURL6T47UEH-4)
See "Search Open" mock-up

Functionality: User clicks on the search icon in the header > search bar appears and field is selected so that users can directly start typing > search gets confirmed by hitting enter OR search bar gets closed by clicking on the closing icon (x)

affabillyty:
Update from Lisa: The closing icon "x" will get a different design, so stay tuned for that update

edlopez000:
Hi @rh9891, we're doing a quick inventory of issue statuses and wanted to see if you had any updates?

gcardonag:
![](https://github.trello.services/images/mini-trello-icon.png) [Create V2 SearchBar](https://trello.com/c/4ydcaDfX/1-create-v2-searchbar)

freelyPheobe:
Hey @RNR1, we see you're assigned to this. Is this something you're actively working on?

freelyPheobe:
Hey @chethtrayen, we haven't seen you in a while. Is there support needed on this ticket?

tomporvaz:
The search bar is currently not working on Phlask mobile. We'd like to make the search appear when the search button is clicked.  An updated design is available here: https://www.figma.com/file/VGGqwl3Eq3GQIAO9I6LoNW/PHLASK?type=design&node-id=8036-117041&mode=design.
Pull request #325 made progress on this ticket, but still contained bugs.

tomporvaz:
Completed in PR #389

lxh5395:
Desktop looks good.

[...]

PR #440, opened 2024-06-06
Fix behavior of map tint due to search bar changes

### LINKED ISSUE #410: Update Mobile Search Overlay Behavior
When using the search feature on mobile, a user should have the ability to deselect the text box, like on desktop, by either pressing enter, selecting one of the search results, or tapping on the map (either pin or background). Deselecting the field would keep the search bar present with any current inputs, but remove the dark overlay over the map.

The dark overlay should be removed when the search bar is deselected on mobile.

https://www.figma.com/file/VGGqwl3Eq3GQIAO9I6LoNW/PHLASK?type=design&node-id=8036-117041&mode=design&t=sIaTQXTFkp2mkx2a-0

n/a

**Additional context**
This search bar for version 2.0 was implemented in issue #228

ravimakhijapi:
The map tint is now deactivated in response to search bar behavior when pressing enter, selecting one of the search results, or tapping on the map (pin or background). Addresses issue #410. 

Related Issues: #410, #428 
Related PRs: #228 

Note to reviewer: _Most_ of the changes in `src/components/SelectedTap/SelectedTap.js` are just cosmetic, e.g. due to `prettier` formatting.


ravimakhijapi:
Hi, the merge conflicts with develop should be addressed now. And, below are a few additional notes for the reviewer. 

- One of the trickier edge cases occurred as follows: if a user clicks the search bar, then clicks on a tap, and then clicks on the search bar, the last click will cause the search bar to become focused (and so the map tint is activated). This seems unnatural however, I think, since the user may have just wanted to go back to the map to look around, not necessarily activate search, thus requiring an additional user click to turn the tint off. Hence, to prevent this issue, an additional piece of redux state was introduced to track this edge case: `tapInfoOpenedWhileSearchOpen`
- After reviewing, I think the `onBlur` event works well here, e.g. it probably doesn't make sense to switch to using `useOnClickOutside` here. 
- Also, made the switch to using `createAction`. 

cc: @gcardonag, @tomporvaz

gcardonag (approved):
Thanks for submitting this fix! 🛠️  It took me a few review attempts to follow the logic behind the edge case you were describing, but I appreciate you accounting for that scenario.

ravimakhijapi:
Ah yes, there were quite a few possibilities to account for, and I was _trying_ to express things in a concise way so as not to give a larger but more confusing amount of detail. Apologies if it could have been clearer. Thanks for reviewing and the merge!

PR #455, opened 2024-06-22
Correct submission of resources

### LINKED ISSUE #391: [V2] Update form submission for new data format
With the new [data format](https://docs.google.com/document/d/1ES2_ZrWGLYzAMOkri9nHrcAsvj5yHz979Y9HKwcFfuY/edit) defined, the submission forms need to be updated to match the new schema. For instance, the address is now optional, where the address was required before.

The first part of this task is to define the new schema as a class, just so we can have some structure. Then, the existing fields in the form need to be updated to match this new type. More specifically, the code at [AddResourceModalV2.jsx - line 231](https://github.com/phlask/phlask-map/blob/develop/src/components/AddResourceModal/AddResourceModalV2.jsx#L231-L302) is where we take the information from the add resource forms, and then submit it to our firebase DB.

Please talk to @vontell for more context. Note that this does not cover adding new information like lat/lon coordinates - that will be covered in other tickets.

vontell:
This PR does a few things:

1. Fixes the submission form to properly track value changes
2. Remove all pages that are no longer used related to submissions
3. Adds better display options for resources (the selected resource modal now shows all information about the resource)
4. The newly added resources are also added to the current data, so a page refetch is not necessary
5. Websites are no longer required


src/components/SelectedTapMobile/SelectedTapDetails.js
```diff
@@ -235,4 +226,36 @@ function SelectedTapDetails(props) {
...
+    tags.push(...(resource.water.dispenser_type || []));
+  }
+  if (resource.food) {
+    tags.push(...(resource.food.food_type || []));
+    tags.push(...(resource.food.distribution_type || []));
+    tags.push(...(resource.food.organization_type || []));
+  }
+  if (resource.forage) {
+    tags.push(resource.forage.forage_type);
+  }
```
gcardonag, line 256: Just to confirm, do bathroom resources not have any resource-specific information? Want to make sure that wasn't an unintended omission

gcardonag: Confirmed that bathrooms do not have resource-specific information

gcardonag (approved):

PR #430, opened 2024-05-09
Refactor resource modal

ravimakhijapi:
# Pull Request

## Change Summary

Integrates the desktop and mobile component code for the Resource modal under a common directory and uses a more consistent approach. Previously, the mobile component was in `src/components/ResourceMenu`, but now both desktop and mobile versions are in `src/components/ChooseResource`.

## Change Reason

It seemed confusing to have the resource modal for desktop and mobile in different places, named differently (e.g. `ChooseResource` vs. `ResourceMenu`), and using different approaches. The hotfix in #421 addressed the mobile resource modal regression, and this new PR provides a more robust solution by refactoring the related code for clarity. 

Related Issue: #422, #401

Issue #401, opened 2024-03-20
MOBILE Resources button


freelyPheobe:
**Describe the bug**
When the resource button is clicked, it doesn't return anything.

**To Reproduce**
1. Go to 'https://phlask.me/'
2. Click on 'resource button on mobile'
3. The image looks nothing like how it should.
4. Compare the screenshots below of the current versus the expected.

**Expected behavior**
![image](https://github.com/phlask/phlask-map/assets/104399179/ddc39b0a-98e8-474a-bcb2-b1a0ffd23f56)
The Resource pages should mimic the above photos including the buttons. The below link is the figma file with the expected design.
https://www.figma.com/file/VGGqwl3Eq3GQIAO9I6LoNW/PHLASK?type=design&node-id=8036-116730&mode=design&t=AHepo2wnzLAQ61J9-0
**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**
Please refer to Design Circle for clarity.

affabillyty:
Hey @freelyPheobe , just a heads up here: the 'design' of the beta.phlask.me toolbar does align with the intended figma designs for PHLASK 2.0. You referenced the current live www.phlask.me, which would be our live PHLASK 1.0.

That said, the issue referenced - the Resources Interface button does not currently work. There is no 'pop up' menu with the 4 Resource icon buttons.

This ticket could be amended to remove the 'old' PHLASK 1.0 screenshots from the live site and focus on the beta site instead.

tomporvaz:
Found commit that broke the resources on mobile, and has fix to test.

tomporvaz:
PR #421 is a hotfix and not intended to resolve this issue.  This issue should remain open until refactoring of toolbar is complete to follow newer code implementation.

lxh5395:
Resource switcher looking good 🙌. Checks off all the boxes in my eyes!!

PR #463, opened 2024-06-25
Refactor resource modal - v2

### LINKED ISSUE #422: Consolidate how the resource modal is rendered on mobile and desktop to potentially use the approach built in PR #332

There appears to be opportunity to consolidate the approaches for how the resource modal is rendered on mobile and desktop. Currently two different approaches are used, which does not seem ideal from a maintenance or code organization point of view. 

In PR #332, the action `TOOLBAR_MODAL_RESOURCE` was added, and is currently being used for the desktop version. This can be seen in component `ChooseResource.js`. 

On the other hand, the mobile approach currently uses action `TOGGLE_RESOURCE_MENU`. This can be seen in component `ResourceMenu.js`. The relevant code for this approach can be seen in PR #421, which might need to be removed after the refactoring. 

The goal is to assess what the best approach for the resource modal is, and use a unified approach for both desktop and mobile. This might entail integrating component `ResourceMenu.js` into component `ChooseResource.js`, however more investigation needs to be done to be sure. 

**Additional context**

Related PRs: #332, #421
Related issues:  #401, #345

Mobile resources menu screenshot:

Resources_Menu_Shown

Desktop resources menu screenshot:

Resources_Menu_Shown-Desktop

ravimakhijapi:
This PR replaces the prior PR #430, and addresses issue #422. 

The desktop and mobile resource selection modals have been refactored as follows:

- Removed the prior mobile approach to show/hide the resource menu and replaced with the desktop approach. Previously, `toggleResourceMenu` was used with state `isResourceMenuShown`. Now, `setToolbarModal` is used and is consistent with how other toolbar modals are shown/hidden.
- Removed the two actions `TOGGLE_RESOURCE_TYPE` and `CHANGE_RESOURCE_TYPE` in favor of a new action called ~~`SET_SELECTED_PLACE`~~ `SET_RESOURCE_TYPE`, which maintains the same functionality, but is named consistent with other actions in `actions.js` and is used for both desktop and mobile. 
- Renamed the mobile component `ResourceMenu` to `MobileChooseResource`, and the analogous desktop component to `DesktopChooseResource`, and moved both to `src/components/ChooseResource`. 

Also, some miscellaneous clean up was done, in particular:

- Remove no longer used `TypeToggle.js` component

Related Issues: #422, #401


gcardonag:
Just a quick comment on the PR description, is it fair to assume that the second bullet meant to say `SET_RESOURCE_TYPE` instead of `SET_SELECTED_PLACE`? I believe that's the intent given the changes but wanted to confirm.

gcardonag:
@ravicodelabs - These changes all look great, thank you for bringing them in! Before approving, there was one thing we can talk about on today's meetup with the wider group. Now that the resource menu on mobile is showing, I think it would be good to run `dispatch(setToolbarModal(TOOLBAR_MODAL_NONE))` when clicking on one of the resource menu options on mobile. My reasoning for doing so differently than on Desktop is that on the desktop you can see the rest of the map and therefore the side-effect of clicking the resource selector. On mobile, the selector takes up most of the screen, so it isn't as clear as a user if my action caused an effect on the map. I believe closing the modal will serve as a response to the click and bring visibility to the pins for the new resource type selection.

ravimakhijapi:
> Just a quick comment on the PR description, is it fair to assume that the second bullet meant to say `SET_RESOURCE_TYPE` instead of `SET_SELECTED_PLACE`? I believe that's the intent given the changes but wanted to confirm.

Yes, this assumption is correct, I've edited the prior comment.

ravimakhijapi:
> ... Now that the resource menu on mobile is showing, I think it would be good to run `dispatch(setToolbarModal(TOOLBAR_MODAL_NONE))` when clicking on one of the resource menu options on mobile...

All makes sense, will work on that change! Also, it looks like the mobile resources menu is broken again (the menu is not showing) on beta.phlask.me, ~~so will take a look at that as well.~~ but was fixed again here in d71fab96e83dc674b23243c7a21289937cc88e5a. 

So for reference, the commit timeline for the resource menu breaking and getting fixing then should be:

1) hot fix done in 6b0a4e8acf9bed1817a0765aa29d305b894fe49f
2) resource menu broke again in 9536e6af967cbc3c6870f988cf6c5bbf73f4b65e
3) fixed again in this PR (but not merged into develop yet) in d71fab96e83dc674b23243c7a21289937cc88e5a while merging develop into the PR branch

ravimakhijapi:
cc: @gcardonag 

The requested update for mobile behavior (to leave the menu open) should be done now. Also, made a few other changes:

- Renamed the components to use "Type" as in `ChooseResourceType`, so that it is not confused with the other `ChooseResource` component in `AddResourceModal/ChooseResource.jsx`
- The data for resource types (icons, text label, etc.) is now being shared among both Desktop and Mobile components. There might be room to do so for `AddResourceModal` as well, but I think that may be out of scope here. 
- Brought up to date with branch develop. However, the build/tests are not working currently as I think we still have to resolve https://github.com/phlask/phlask-map/pull/506.

gcardonag (approved):
Thanks for making all of those changes! Approved ✅

PR #501, opened 2024-07-24
Fix map snapback by cleaning up hooks

### LINKED ISSUE #459: Avoid panning to user location in most cases
When clicking off of a resource detailed modal view, or doing really any operation on the website, the map will often pan back to your location. We should make it so that this panning operation only happens when you first open the site - otherwise, the map should not pan to the user location.

### LINKED ISSUE #500: Fix bug where map doesn't follow user.
**Describe the bug**
The map at beta.phlask.me does not follow the user, resulting in a subpar experience. If you click on a resource then close it, your viewport is set back to City Hall. This also happens in the search functionality to a smaller degree, where you're set back to the search pin.

I think this is a bug because there are multiple conflicting statements in the useEffect hook in `ReactGoogleMaps.js` and the initialCenter and center params in the Map component.

**To Reproduce**
1. Go to the map
2. Click on any given resource away from City Hall
3. Click off the resource or close the pop-up
4. See the map reset your viewport to City Hall

**Expected behavior**
If the user closes out of a resource or taps away from the resource pop-up, the map should stay at the location of the last viewed resource.

Occurs on mobile & desktop.

**Additional context**

cjfit:
# Pull Request

## Change Summary

Fixes the map snapping back to City Hall when browsing resources. Map will now follow user, including in the search feature.

## Change Reason

Map snapping back to City Hall marks for a poor user experience when trying to browse resources on the map. It also conflicts with the panning behavior of the search functionality.

## Verification [Optional]

[phlask-500-480p.mov.zip](https://github.com/user-attachments/files/16355713/phlask-500-480p.mov.zip)

No new console errors:
Screen Shot 2024-07-23 at 8 41 04 PM

Fixes #500

### DIFF
--- a/src/components/ReactGoogleMaps/ReactGoogleMaps.js
+++ b/src/components/ReactGoogleMaps/ReactGoogleMaps.js
@@ -162,8 +162,8 @@ export const ReactGoogleMaps = ({ google }) => {
   );
   const toolbarModal = useSelector(state => state.filterMarkers.toolbarModal);
 
-  const [currentLat, setCurrentLat] = useState(mapCenter.lat);
-  const [currentLon, setCurrentLon] = useState(mapCenter.lng);
+  const [currentLat, setCurrentLat] = useState(CITY_HALL_COORDINATES.latitude);
+  const [currentLon, setCurrentLon] = useState(CITY_HALL_COORDINATES.longitude);
   const [zoom, setZoom] = useState(16);
   const [searchedTap, setSearchedTap] = useState(null);
   const [map, setMap] = useState(null);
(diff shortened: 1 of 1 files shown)


gcardonag:
Reviewing this on my phone so I'm a bit limited, but it looks like this would eliminate the map using the user's location to center the map where they are. Could you keep that functionality as a part of the fix? Users may just want to see options near them when browsing the map on their phones. It may be confusing to see city hall on the map when you're browsing from North Philly.

For some more context, we use City Hall as a default first location in the event users do not want to share their location.

cjfit:
Thanks for the catch @gcardonag! I wasn't sure if the geolocation feature was implemented yet. The latest commit adds that back.

src/components/ReactGoogleMaps/ReactGoogleMaps.js
```diff
@@ -181,39 +181,19 @@ export const ReactGoogleMaps = ({ google }) => {
...
-    const setDefaultLocation = () => {
-      setCurrentLat(parseFloat(CITY_HALL_COORDINATES.latitude));
-      setCurrentLon(parseFloat(CITY_HALL_COORDINATES.longitude));
+    const fetchCoordinates = async () => {
+      try {
+        const position = await getCoordinates();
+        setCurrentLat(position.coords.latitude);
+        setCurrentLon(position.coords.longitude);
+      } catch (error) {
+        console.error("Error obtaining geolocation. Default coordinates will be used.");
```
RNR1: Can we noop this catch statement? We don't get much value from these console statements

src/components/ReactGoogleMaps/ReactGoogleMaps.js
```diff
@@ -181,39 +181,19 @@ export const ReactGoogleMaps = ({ google }) => {
...
     };
-    getCoordinates()
-      .then(position => {
-        if (
-          Number.isNaN(position.coords.latitude) ||
-          Number.isNaN(position.coords.longitude)
-        ) {
-          setDefaultLocation();
-        } else {
-          dispatch(
```
RNR1, line 196: I noticed we removed these setters. are they not necessary anymore?

RNR1 (commented):
A few questions/suggestions

src/components/ReactGoogleMaps/ReactGoogleMaps.js
cjfit: Correct

src/components/ReactGoogleMaps/ReactGoogleMaps.js
cjfit: If we yeet the catch, we'll get an unhandled GeolocationPositionError from `getCoordinates`. I removed the console statement in the latest commit though
Screen Shot 2024-08-11 at 7 59 16 PM

gcardonag (approved):
Approved since the logging message has been cleaned up and the rest of the code has been reviewed ✅

PR #506, opened 2024-08-01
Hotfix for broken selector import

ravimakhijapi:
There seems to be a broken import which is causing the app not to run in a local (docker) development environment (though notably, the app is still working at beta.phlask.me, ~~so perhaps there is a difference elsewhere, such as in the production build/server approach~~, I think the beta.phlask.me app may be falling back to the last successful build). This PR fixes that import and allows the app to start normally in a local development environment again.

### DIFF
--- a/src/components/Toolbar/Toolbar.js
+++ b/src/components/Toolbar/Toolbar.js
@@ -41,7 +41,7 @@ import { resourceTypeSelector } from '../../selectors/filterMarkersSelectors';
 import ResourceMenu from '../ResourceMenu/ResourceMenu';
 import NavigationItem from './NavigationItem';
 import useIsMobile from 'hooks/useIsMobile';
-import selectFilteredResource from '../../selectors/waterSelectors';
+import selectFilteredResource from 'selectors/resourceSelectors';
 
 // Actual Magic: https://stackoverflow.com/a/41337005
 // Distance calculates the distance between two lat/lon pairs
@@ -53,7 +53,7 @@ function distance(lat1, lon1, lat2, lon2) {
     (Math.cos(lat1 * p) *
       Math.cos(lat2 * p) *
       (1 - Math.cos((lon2 - lon1) * p))) /
-      2;
+    2;
   return 12742 * Math.asin(Math.sqrt(a));
 }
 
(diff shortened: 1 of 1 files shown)


src/components/Toolbar/Toolbar.js
```diff
@@ -41,7 +41,7 @@ import { resourceTypeSelector } from '../../selectors/filterMarkersSelectors';
 import ResourceMenu from '../ResourceMenu/ResourceMenu';
 import NavigationItem from './NavigationItem';
 import useIsMobile from 'hooks/useIsMobile';
-import selectFilteredResource from '../../selectors/waterSelectors';
+import selectFilteredResource from 'selectors/resourceSelectors';
```
vontell: What is the error you are getting in the docker container? I don't think this is correct because this will attempt to search for a npm-installed package of this name, rather than from a local file.

vontell (changes requested):

src/components/Toolbar/Toolbar.js
ravimakhijapi: I believe the import (after my change) was resolving to the local module. However, even if that is correct, relative imports are probably better/safer here, true, so I have added a commit to do that. I also went ahead and updated the previous import in the code, which was also not using "dot notation" relative imports. Please let me know if I've missed something. 

I am currently on the go so I will have to recheck what the exact error was in a bit, but I believe the error pointed out the broken import (for the water selector file, which I believe was removed).

ravimakhijapi: Here is a screenshot of the exact error.

image

ravimakhijapi: > ...this will attempt to search for a npm-installed package of this name, rather than from a local file.

I think that's right for node's resolution process. On the other hand, we are currently using webpack via react-scripts for the build, which has its own resolution process, and I am guessing that's why the import worked even without a "dot" relative import. Though, I am not sure what the default webpack config used by react-scripts is, so I would have to dig a bit more to confirm this.

gcardonag (approved):
Approved, this will fix the issue breaking the latest builds in the develop branch. Thanks @ravicodelabs !

gcardonag:
Going to merge this since the requested changes have already been addressed.

PR #507, opened 2024-08-06
Set Phlask DB dynamically based on environment variable

### LINKED ISSUE #498: The environment should properly switch to the correct database
Currently, all environments will use the `phlask-beta` database. We should:

* Create the `phlask-test` and `phlask-prod` databases by cloning the `phlask-beta` dastabase
* The frontend should properly choose the right DB to use if it is in prod vs local vs beta.

cjfit:
# Pull Request

## Change Summary

- A simple one-liner added to `firebaseConfig.json` that sets the Firebase DB URL based on the presence of an environment variable `REACT_APP_DB_URL`. If `REACT_APP_DB_URL` is not present, it will default to the test database URL.
- The Github action that builds the Docker image was updated to use `REACT_APP_DB_URL`. 
- Unused `utils.js` file removed

## Change Reason

Contributors to the project are typically testing, and any changes made (e.g. adding a resource) will be reflected on the production database. This is not ideal and leads to a lot of data cleanup actions. This change ensures the test database will be used by default unless the developer wants to force use the prod DB.

## Testing
Tested locally the environment variable override works and by resources retrieved/added are done to to the proper DB. 

For the GitHub Action, we'll want to deploy the beta site and then confirm resource changes are made to the production DB.

## Verification [Optional]

Closes #498


cjfit:
Closing to re-open anew

cjfit:
changes in https://github.com/phlask/phlask-map/pull/508

PR #508, opened 2024-08-06
Set Phlask DB dynamically

### LINKED ISSUE #498: The environment should properly switch to the correct database

cjfit:
# Pull Request

## Change Summary

- A simple one-liner added to `firebaseConfig.json` that sets the Firebase DB URL based on the presence of an environment variable `REACT_APP_DB_URL`. If `REACT_APP_DB_URL` is not present, it will default to the test database URL.
- The Github action that builds the Docker image was updated to use `REACT_APP_DB_URL`. 
- Unused `utils.js` file removed

## Change Reason

Contributors to the project are typically testing, and any changes made (e.g. adding a resource) will be reflected on the production database. This is not ideal and leads to a lot of data cleanup actions. This change ensures the test database will be used by default unless the developer wants to force use the prod DB.

## Testing
Tested locally the environment variable override works and by resources retrieved/added are done to to the proper DB. 

For the GitHub Action, we'll want to deploy the beta site and then confirm resource changes are made to the production DB.

## Verification [Optional]

Closes #498

### DIFF
--- a/.github/workflows/betasite.yml
+++ b/.github/workflows/betasite.yml
@@ -6,7 +6,7 @@ on:
       - develop
 env:
   AWS_DEFAULT_REGION: us-east-2
-  REACT_APP_DB_URL: https://phlask-prod.firebaseio.com
+  REACT_APP_DB_URL: https://phlask-beta.firebaseio.com
 permissions:
   id-token: write   # This is required for requesting the JWT
   contents: read    # This is required for actions/checkout
--- a/.github/workflows/betasite.yml
+++ b/.github/workflows/betasite.yml
@@ -6,6 +6,7 @@ on:
       - develop
 env:
   AWS_DEFAULT_REGION: us-east-2
+  REACT_APP_DB_URL: https://phlask-prod.firebaseio.com
 permissions:
   id-token: write   # This is required for requesting the JWT
   contents: read    # This is required for actions/checkout
--- a/src/components/AddResourceModal/utils.js
+++ b/src/components/AddResourceModal/utils.js
@@ -1,47 +1,9 @@
-import { initializeApp } from 'firebase/app';
-
-export const connectToFirebase = (hostname, resourceType) => {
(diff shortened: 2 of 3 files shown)


vontell (changes requested):
@gcardonag the current site is only hosted via s3 correct? We were wondering if there is a way to set env vars there.

If the answer is no, @cjfit you can probably use the existing code in utils.js to decide the environment, and then use that inside of the firebaseConfig.js to select the right DB.

cjfit:
I think env vars may be embedded into the React build - once the action is fixed we could verify this

gcardonag:
@cjfit is correct, the environment variables will be resolved during the build phase, so the time to set it would be in GitHub actions. The resolved value would be preserved in `bundle.js` during the upload to S3.

Source: https://create-react-app.dev/docs/adding-custom-environment-variables/ 

I tested it locally and got two different lines in `bundle.js` depending on whether the environment variable is set:
- `yarn start` -> `let phlaskDatabaseUrl = ({"NODE_ENV":"development","PUBLIC_URL":"","WDS_SOCKET_HOST":undefined,"WDS_SOCKET_PATH":undefined,"WDS_SOCKET_PORT":undefined,"FAST_REFRESH":true}).REACT_APP_DB_URL || "https://phlask-test.firebaseio.com";`
- `REACT_APP_DB_URL="teststring" yarn start` -> `let phlaskDatabaseUrl = "teststring" || 0;`

gcardonag (approved):
Approving this and merging as the only test failure is a one-off due to the switch from docker-compose v1 to docker-compose v2 for GitHub Actions runners and is not directly related to this PR.

PR #514, opened 2024-08-07
Update devcontainer.json

ravimakhijapi:
Updates the devcontainer.json file for both GitHub Codespaces as well as local VS Code docker setups (when devcontainer.json is being used). This shouldn't change much for folks using these tools already, but some improvements are made on the devcontainer.json.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,23 @@
+ARG VARIANT=dev-16-bookworm
+FROM mcr.microsoft.com/devcontainers/javascript-node:${VARIANT}
+
+ARG USERNAME=node
+ARG NPM_GLOBAL=/usr/local/share/npm-global
+
+# Add NPM global to PATH.
+ENV PATH=${NPM_GLOBAL}/bin:${PATH}
+
+RUN \
+    # Configure global npm install location, use group to adapt to UID/GID changes
+    if ! cat /etc/group | grep -e "^npm:" > /dev/null 2>&1; then groupadd -r npm; fi \
+    && usermod -a -G npm ${USERNAME} \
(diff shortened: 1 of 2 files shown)


gcardonag (approved):
Approved ✅ 
Before merging, and for the benefit of folks who may review this PR in the future, what was the goal of these improvements?

The startup speed seems similar at a glance, but I recall talking to you about improvements for local dev where it may have not been working.

ravimakhijapi:
> Approved ✅ 
> 
> Before merging, and for the benefit of folks who may review this PR in the future, what was the goal of these improvements?
> 
> 
> 
> The startup speed seems similar at a glance, but I recall talking to you about improvements for local dev where it may have not been working.

Ah yes, the main goal here was to address an issue where the prior image would not work on an M processor based MacBook, as their was a conflict in the 'universal' image which did not allow the interoperability. 

Regarding the fix: the configuration was updated to use a 'javascript-node' image rather than 'universal', which doesn't have the problematic dependency, and node is sufficient for our code base.

gcardonag:
> > Approved ✅ 
> > 
> > Before merging, and for the benefit of folks who may review this PR in the future, what was the goal of these improvements?
> > 
> > 
> > 
> > The startup speed seems similar at a glance, but I recall talking to you about improvements for local dev where it may have not been working.
> 
> Ah yes, the main goal here was to address an issue where the prior image would not work on an M processor based MacBook, as their was a conflict in the 'universal' image which did not allow the interoperability. 
> 
> Regarding the fix: the configuration was updated to use a 'javascript-node' image rather than 'universal', which doesn't have the problematic dependency, and node is sufficient for our code base. 

Thank you @ravicodelabs ! 😄

PR #516, opened 2024-08-10
Add Mobile Filter and Change Mobile Add Resource

### LINKED ISSUE #511: filter menu on mobile
A swipable drawer should be implemented for filtering resources on mobile.  Filter logic has already been implemented, but this feature also needs to be available on mobile.  Follow these figma designs: https://www.figma.com/design/VGGqwl3Eq3GQIAO9I6LoNW/PHLASK?node-id=8036-111314&m=dev

astoppop:
# Pull Request

## Change Summary

Changed mobile add resource design, along with some changes to make it more consistent with choose resource. Implemented filter for mobile.

## Change Reason

Mobile add resource design didn't match Figma. Mobile filter was not implemented.

Related Issue: fixes #511


gcardonag (changes requested):
It seems a few of our tests are failing when applied to some of these changes. I tried digging into the cause of the issues, and here are a few things to address:

- The test at `cypress/e2e/desktop/crowdsourcing.cy.js` normally targets an element with `data-cy=button-contribute-water` element. This code replaced it with an element that has `data-cy=button-WATER-data-selector`
  - I tried to update the test to replace the test targeting to use the new naming convention, but it fails because the site is rendering two elements with the same value of `data-cy=button-WATER-data-selector`. This should probably be fixed since there should only be one of these rendering based on whether the site is mobile or desktop.

gcardonag (approved):
Approving and merging this as I will apply a separate commit to address the issues in my comment

PR #528, opened 2024-09-05
docs: typo fix in README.md - use use

waltribeiro:
# Pull Request

## Change Summary

`use use` in [Line 54](https://github.com/phlask/phlask-map/blob/c9412429d0194f76edb92e4c568359d5076b3b08/README.md?plain=1#L54)

## Change Reason

Very important crucial bug fix. Major revision.


gcardonag (approved):
Thank you for this @waltribeiro ! Every change makes a difference 🦾

PR #535, opened 2024-09-24
[Issue 504] Allow undefined address components, and fix weird path issue

### LINKED ISSUE #504: When adding a resource, the address does not always parse correctly
The address parsing code assumes that the city, state, and postal code will exist. Update the code to handle cases where this is not true.

While we are working on this, let's also fix an issue where adding resource returns an error regarding "ID 0"

vontell:
* Allows addresses to not return/have certain component such as zip codes, cities, or states
* Fixes a weird behavior in Firebase where the pieces of an ID can show up under a different field sometimes


gcardonag (approved):

gcardonag:
@vontell was able to test this with a location that prompted the issue due to a lack of a zip code on Google (Passerelle du College, Lyon, France) and confirmed this is now working!

Issue #527, opened 2024-09-04
Panning after opening modal pans map to last location


tomporvaz:
**Describe the bug**
In desktop mode (and presumably mobile too) the map will pan to the last location visited specifically after opening the modal and not closing it.  

**To Reproduce**
1. Choose a location (e.g. Franklin Square), click the pin opening the details modal. 
2. Without closing the modal, manually pan the map to another location (e.g. Franklin D. Roosevelt Park).
3. Click the new location (e.g. Franklin D. Roosevelt Park).
4. See error, the map has panned back to the first location (e.g. Franklin Square).

**Expected behavior**
I expect the map to pan to the new location, e.g. Franklin D. Roosevelt Park.

tomporvaz:
Also, when modal is opened and then closed, and then near me button is clicked, no pan. 
Also, if you open a modal, and then click near me, and then exit modal, the map will pan back to last location.

ravimakhijapi:
Hey. Just started by trying to reproduce the stated issues. There are a few notable points. Here are the findings:

- **Issue 1:** Open a location modal, pan to somewhere else (keeping the mentioned modal open), click a new location pin (so that a new modal opens), and the map will incorrectly pan back to the first location. 
	- Desktop - Issue **not** occurring for me in Arc, Chrome, or Safari. 
	- Mobile - Not applicable to mobile - can't pan the map while a modal is open. 
- **Issue 2** - Near Me not panning to current location after opening/closing a modal. 
   - Reproduced issue in Desktop. 
   - Reproduced issue in Mobile. 
- **Issue 3** - If you open a modal, and then click near me, and then exit the modal, the map will pan back to last location
  - Desktop - Issue seems present, but might be more nuanced. In particular, when location access is _not_ given to the browser, it seems to go back to the last location as previously stated. However, when location access is given to the browser, it is going back to the current location I believe. 
  - Mobile - Not applicable - the _Near Me_ button is hidden when modal is open, I think. 

Note also that at the moment, I believe current location would be the default current location (around city hall I believe) when location services are off.

ravimakhijapi:
Hi. I'm leaving some of my testing notes here, which organizes the various cases, in case someone else finds it helpful in debugging. 

<google-sheets-html-origin>
Issue and Environment | Location On | Location Off | Issue Details
-- | -- | -- | --
[...]

PR #539, opened 2024-10-06
Fix map panning issues related resource info modal

ravimakhijapi:
## Change Summary

Addresses #527. Introduces `lastResourcePan` in redux state to supersede prior `mapCenter` in redux, and `currentLat/currentLon` component state in `<ReactGoogleMaps>`, which helps properly track the last pan caused by the user clicking a pin or on the "Near Me" button. 

## Change Reason

Fixes various issues as described in detail in issue #527, where the map _was_ panning when it shouldn't be, and _wasn't_ panning when it should be, depending on the scenario.

Related Issue: #527


src/reducers/filterMarkers.js
```diff
@@ -129,7 +130,7 @@ export default (state = initialState, act) => {
         : {
           ...state,
           selectedPlace: state.allResources[act.selectedPlace],
-          showingInfoWindow: true
+          showingInfoWindow: true // debuggin! why is this here, and not in the other branch?
```
gcardonag: Just checking, does this comment still need to be addressed or can it be removed?

gcardonag (commented):
Overall, the changes look good and I was able to retest with your detailed steps, thanks again for writing them! 😄 

I only found one small detail to address before approving

src/reducers/filterMarkers.js
ravimakhijapi: Ack (no pun intended!) Thanks for catching this, it was a self-comment while I was working, which has now been removed. 

To clarify, by "branch" I meant the `if` branch in `filterReducers`, which did not set the selected place, while the `else` branch did. After a bit more digging, I could not find any instance where we are actually using the resource ID to set the place, hence the if-else seemed unnecessary, removing it altogether removed the need to worry about setting the selected place in both branches. 

TL;DR - I have removed the if-else and that comment in the reducer, and also transitioned to the `createAction` syntax as part of the recent guidelines. Also, some of the commit changes are being caused by prettier formatting, sorry for any confusion!

ravimakhijapi:
I believe your last comment was addressed. Please let me know if I've missed anything!

gcardonag:
> I believe your last comment was addressed. Please let me know if I've missed anything!

Nope, your change makes sense! I'm thinking the check for typeof in the old `SET_SELECTED_PLACE` action was tied to code that is no longer in-place today, so removing it makes sense.

gcardonag (approved):

PR #540, opened 2024-10-15
Make sure to treat forage_type as a list

vontell:
This was a bug that Billy found when adding a foraging site. It turns out there was an issue where we were trying to show the foraging_type as a string when it was actually a list of possible options.


gcardonag (approved):
Approved, thanks for the quick fix! I'll deal with the flaky test failure on a separate branch

Issue #460, opened 2024-06-23
Catch errors when converting address to lat/lng


vontell:
In some cases, attempting to create a new resource results in an error message. Sometimes this is related to a failure in geocoding the address (i.e. trying to get the lat/lng for an address), but in some other cases, another error happens. We should capture any error when submitting a resource and display a message in the dialog, rather than crashing the app.

gcardonag:
@vontell , I know we shared this during today's standup, but I'll look for your help to confirm whether this is still active while I review other tickets tonight. 🙏

PR #549, opened 2024-11-30
[Issue 460] Catches goecoding issue before reaching submission

vontell:
# Pull Request

## Change Summary

This PR fixes a bug where addresses that aren't real or don't have a valid lat/lng would cause the app to crash without any indication to the user.

This change "fixes" it by catching this issue before we even get to submission - if the address is not an address that would result in a latitude and longitude, an error is shown saying that it is not a valid address.

Screenshot 2024-11-30 at 4 39 00 PM

## Verification

Here is a video of this being demonstrated: https://www.loom.com/share/a0964ad0e4e7443491f9f0ce09cd1a87?sid=1467be47-5d0f-4e02-be58-e4642604561c

Related Issue: #460

### DIFF
--- a/src/components/AddResourceModal/AddBathroom/AddBathroom.jsx
+++ b/src/components/AddResourceModal/AddBathroom/AddBathroom.jsx
@@ -33,7 +33,8 @@ function AddBathroom({
   handicapAccessible,
   hasFountain,
   checkboxChangeHandler,
-  textFieldChangeHandler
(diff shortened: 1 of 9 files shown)


gcardonag (approved):
Nice! I think showing the error sooner makes sense. I tested manually typing in a valid address instead of using autocomplete and can confirm the validation still works properly ✅

Issue #550, opened 2024-12-11
Keep track of location permissions inside app state


GV14982:
**Is your feature request related to a problem? Please describe.**
Right now if you deny the app location access, it seems to just assume you're at city hall. This conflicts with the new "Near Me" feature.

**Describe the solution you'd like**
We should add some logic to keep track of whether or not we have access to a user's location. This would allow us to dynamically enable/disable features based on that data to reduce undefined behavior in the app.

**Describe alternatives you've considered**

**Additional context**
Spoke with @affabillyty at the December Hack Night and this seems to be a pretty pressing concern. This will require a follow up ticket specifically for the "Near Me" button to be disabled based on location status.

GV14982:
One limitation of my current implementation of this is that Firefox "temporary permissions" don't change the permission state like other browsers, so if a user doesn't check the "Remember this option" box on the prompt, it will be stuck in a sort of limbo.

Bugzilla ticket for reference: https://bugzilla.mozilla.org/show_bug.cgi?id=1719651

I would love some suggestions for if I should even worry about this (I'm probably an outlier in that I generally use temp perms).

@affabillyty @gcardonag @RNR1

gcardonag:
I know we talked about this on standup, so capturing it here for future reference. It seems like we should explore this from a user experience perspective.

In my opinion, this is definitely more of a good-to-have than a need-to-have, so no need to suffer too much if this becomes difficult to explore. I generally appreciate Firefox's approach to location sharing as it helps me when testing to use our default location

PR #556, opened 2025-01-08
feat: add hook to get users geolocation api status

GV14982:
# Pull Request

## Change Summary

- Added new global redux reducers for handling user state
- Setup logic for tracking a users `geolocation` permissions and updating the redux store
- Built small utility hook for getting that info easily from the store

## Change Reason

We can use this data to disable features that might require
location information.

## Verification [Optional]

A .gif, video, or screenshots of the feature are extremely helpful in verifying that code changes are working correctly. Feel free to include them in the PR to help the reviewer.

We recommend using [OBS](https://obsproject.com/) to record videos, as it is open source, free, and available on all major operating systems.

Related Issue: #550


gcardonag:
@GV14982 @RNR1 , I gave this PR a look. The merge conflict is just a result of the larger PR that moved us to Vite, including a move from `src/App.js` to `src/App.jsx`. I tried locally merging the `develop` branch in and then bringing the `App.js`-specific code back in, but I had to make a change because we moved the redux `<Provider>` into this `App` component.

I refactored it by breaking out the Geolocation Interval checker into another component in the same file (see below). @RNR1 , do you think that approach is ok or should we break it out into a totally separate component file to keep `App.jsx` cleaner?
```typescript
import { CssBaseline } from '@mui/material';
import { ThemeProvider } from '@mui/material/styles';
import { useEffect, useCallback } from 'react';
import './App.css';
import { initAnalytics, logPageView } from 'analytics';
import Head from 'components/Head/Head';
import MapPage from 'components/MapPage/MapPage';
import theme from 'theme';
import store from 'store';
import { Provider, useDispatch } from 'react-redux';
import { userDataSlice } from 'reducers/user';

/**
 * Handle checking geolocation on an interval.
 *
 * @param {NodeJS.Interval} interval Interval to clear on callback run.
 * @param {(status: boolean) => void} setter Setter to run when the callbacks run.
 */
function geolocationInterval(interval, setter) {
  navigator.geolocation.getCurrentPosition(
    () => {
      setter(true);
      clearInterval(interval);
    },
    () => {
      setter(false);
      clearInterval(interval);
    }
  );
}

/**
 * Check geolocation perms and set an onchange handler.
 * @async
 * @param {(status: boolean) => void} setter Setter to run based on initial query.
 */
async function queryGeolocationPerms(setter) {
  const perms = await navigator.permissions.query({ name: 'geolocation' });
  setter(perms.state === 'granted');
  perms.addEventListener('change', () => {
    setter(perms.state === 'granted');
  });
}

const GeolocationHandler = () => {
  const dispatch = useDispatch();
  const setEnabled = useCallback(
    payload => {
      dispatch(userDataSlice.actions.updateUserLocationEnabled(payload));
    },
    [dispatch]
  );
  useEffect(() => {
    let interval;
    /**
     * This one is a doosy...
     * Basically Firefox doesn't think temporary permissions should trigger the Permissions API onchange event (Which is dumb).
     * This is a workaround that periodically checks for a users location until they respond to the prompt, when it cancels the interval.
     *
     * The main downside here is that we are constantly triggering the access prompt.
     * There is an edge case where a user might click close to when the interval reruns.
     * This will reprompt and make it seem like the prompt didn't close properly.
     *
     * Here is a bugzilla ticket relating to this: https://bugzilla.mozilla.org/show_bug.cgi?id=1719651
     * Hopefully Mozilla changes their mind and will properly handle temporary permissions
     */
    // TODO: Revisit this logic if Mozilla comes to their senses
    if (navigator.userAgent.includes('Firefox')) {
      interval = setInterval(() => {
        geolocationInterval(interval, setEnabled);
      }, 2000);
      geolocationInterval(interval, setEnabled);
      // This is the sensible way to handle it...
    } else {
      queryGeolocationPerms(setEnabled);
    }
    return () => {
      clearInterval(interval);
    };
  }, [setEnabled]);
}

const App = () => {
  initAnalytics();
  logPageView();
  return (
    <Provider store={store}>
      <ThemeProvider theme={theme}>
        <CssBaseline />
        <div className="page-wrapper">
          <GeolocationHandler />
          <Head />
          <MapPage />
        </div>
      </ThemeProvider>
    </Provider>
  );
};

export default App;
```

src/App.js
```diff
@@ -1,16 +1,86 @@
...
+import { useDispatch } from 'react-redux';
+import { userDataSlice } from 'reducers/user';
+
+/**
+ * Handle checking geolocation on an interval.
+ *
+ * @param {NodeJS.Interval} interval Interval to clear on callback run.
+ * @param {(status: boolean) => void} setter Setter to run when the callbacks run.
+ */
+function geolocationInterval(interval, setter) {
```
RNR1: We use this function in one place, and it knows too much about its dependencies. I think that using that logic inline would make it more readable and easier to understand, rather than abstracting it from its original flow

src/App.js
```diff
@@ -1,16 +1,86 @@
...
+    }
+  );
+}
+
+/**
+ * Check geolocation perms and set an onchange handler.
+ * @async
+ * @param {(status: boolean) => void} setter Setter to run based on initial query.
+ */
+async function queryGeolocationPerms(setter) {
```
RNR1: 1. Can we use `Permissions` instead of `Perms`? 
2. Can we rename `setter` to something more self explanatory?

src/App.js
```diff
@@ -1,16 +1,86 @@
...
+  perms.addEventListener('change', () => {
+    setter(perms.state === 'granted');
+  });
+}
 
 function App(props) {
   initAnalytics();
   logPageView();
+  const dispatch = useDispatch();
+  const setEnabled = useCallback(
```
RNR1: nit: for better scoping, can we declare this function inside the useEffect? Also, can we name it to be more specific? maybe `onUserLocationStatusChanged`

src/App.js
```diff
@@ -1,16 +1,86 @@
...
+  setter(perms.state === 'granted');
+  perms.addEventListener('change', () => {
+    setter(perms.state === 'granted');
+  });
+}
 
 function App(props) {
   initAnalytics();
   logPageView();
+  const dispatch = useDispatch();
```
RNR1: since this is running on the app top level, maybe it's worth refactoring it into a hook like `useUserLocationStatus`

src/hooks/useGeolocation.js
```diff
@@ -0,0 +1,11 @@
+import { useSelector } from 'react-redux';
+
+/**
+ * Hook to track the geolocation status in the browser.
+ *
+ * @returns {boolean} Geolocation enabled flag.
+ */
+export default function useGeolocationEnabled() {
+  const enabled = useSelector(state => state.userData.locationEnabled);
```
RNR1: I think it will be more valuable to convert this to a selector function rather than a hook that explicitly uses `useSelector`

RNR1 (commented):
Looks good! A few suggestions 😇

src/hooks/useGeolocation.js
GV14982: That's fair, I had initially planned to put a ton of logic inside the hook, but then moved away from that idea.

src/App.js
GV14982: I think my initial reasoning was to reduce duplication.

RNR1 (approved):
LGTM!

Issue #563, opened 2025-01-16
Migrate data from Firebase to Supabase


vontell:
Migrate the data on Firebase to a Supabase DB. This task also involves connecting it to the frontend and setting up a workflow for granting permissions to the Supabase instance.

PR #567, opened 2025-01-24
Replace Firebase with Supabase

vontell:
Addresses ticket #563 

This PR officially replaces all Firebase functionality from PHLask and instead replaces it with Supabase. The transition was actually quite easy, since the Supabase SDK is quite easy to use!

To check out the data, message me (@vontell) to get access to the project on Supabase. Otherwise, you should be able to pull this branch and see that the behavior is the same as before.

We should keep the Firebase database around for now to verify that things are looking good after testing. After a while, we can officially close down that project :) 

Note that the new Supabase account has two database - the `resources` database, and the `contributors` database.

The one question I have for reviewers - I'd like to insert the API key (don't worry this is a public key similar to the Firebase ones we had) as an environment variable rather than plaintext. Before I do that, I wanted to verify where all the places are that I need to set this. I notice we are not consistent with setting the DB URL, for instance.

Note that part of the changes here seem to be formatting from my editor, sorry for that...

Screenshot 2025-01-24 at 12 36 37 PM
Screenshot 2025-01-24 at 12 40 11 PM
Screenshot 2025-01-24 at 12 40 57 PM


src/db.js
```diff
@@ -0,0 +1,66 @@
+import { createClient } from '@supabase/supabase-js';
+
+// Need access to the database? Message us in the #phlask-data channel on Slack
+const databaseUrl =
+    import.meta.env.VITE_DB_URL || 'https://wantycfbnzzocsbthqzs.supabase.co';
+const resourceDatabaseName =
+    import.meta.env.VITE_DB_NAME || 'resources';
+const databaseApiKey =
+    import.meta.env.VITE_DB_API_KEY || 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6IndhbnR5Y2Zibnp6b2NzYnRocXpzIiwicm9sZSI6ImFub24iLCJpYXQiOjE3MzcwNDY2OTgsImV4cCI6MjA1MjYyMjY5OH0.yczsMOx3Y-zsWu-GjYEajIb0yw9fYWEIUglmmfM1zCY';
```
gcardonag, line 9: Following up on our standup chat, lets create a sample `.env` file for developers to use in the future so that we can pivot away from storing the key in this file.

We should also update the README to include how to request access to the developer `.env` file in the future.

gcardonag (approved):
Thanks for this! Hopefully this'll make our data work down the line a little easier 🤞

PR #576, opened 2025-02-26
fix resources mobile

RNR1:
# Pull Request

## Change Summary

Fix resources modal on mobile

## Change Reason

We want to change resources on mobile

## Verification

image


gcardonag (approved):
Thanks for getting this fix in! 🔧

PR #582, opened 2025-03-11
Adds tests to validate clicking taps does not crash site

gcardonag:
# Pull Request

## Change Summary

Adds a scheduled test to click on each tap and report any failures that cause the page to crash.

## Change Reason

We have recently had a few instances of cases where clicking individual taps has led the site to crash. I believe this test, run on a scheduled basis, can help us minimize the likelihood of these issues, which would be hard to chase by individual users.

## Verification [Optional]

TBD

Related Issue: N/A

### DIFF
--- a/cypress/integration/desktop/tapcheck.bathroom.cy.js
+++ b/cypress/integration/desktop/tapcheck.bathroom.cy.js
@@ -1,0 +1,76 @@
+// Primary Goal: Test each tap in without incurring heavy use of the Maps API, such as loading dynamic maps multiple times
+
+// - Builds a queue of all RESOURCE_TYPE locations
+// - While the queue is not empty, pulls an item from the queue and tests clicking it
+// - If the site crashes, records the tap that crashed, reloads the site, and continues looping
+
+const RESOURCE_TYPE = "BATHROOM"
+
+let exception_ocurred = false;
+let exception_message = "";
+
+// For each resource type, test each site detail permutation and confirm only the expected number of taps appear.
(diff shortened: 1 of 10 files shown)


cypress/integration/desktop/tapcheck.bathroom.cy.js
```diff
@@ -0,0 +1,76 @@
...
+    cy.on('uncaught:exception', (err, runnable) => {
+      exception_message = err.message;
+      exception_ocurred = true;
+      return false;
+    })
+
+    const testLocation = () => {
+      const target_location_id = test_target_queue.pop()
+      cy.task('log', `Testing ${target_location_id}`)
+      cy.get(`[title=data-cy-${target_location_id}]`).click({ force: true, timeout: 8000 }).then(() => {
```
RNR1, line 31: Are we able to use a data-test-id instead of using title attributes? The title attribute is not intended for testing purposes

cypress/integration/desktop/tapcheck.water.cy.js
RNR1, line 1: Are we able to create a reusable function that we can use for each tap rather than repeating the same pattern over 4 different files?

cypress/integration/desktop/tapcheck.bathroom.cy.js
gcardonag: We reviewed this on our weekly call, this seems to be a limitation of the Google Maps Marker Javascript class, which doesn't allow adding custom attributes.

Issue #575, opened 2025-02-19
Crowdsourcing multi-select dropdowns inaccessible at different breakpoints


affabillyty:
**Describe the bug**
The multi-select dropdowns in the crowdsourcing tool (e.g., water dispensers select, food distribution type, etc.) 
When you go to pull them down, they resize the modal and cause it to overflow. This makes it inaccessible for the user to click/access parts that are overflowing beyond the window.

**To Reproduce**
1. On a laptop screen (height: ~680px), click on "Add Site" from the toolbar
2. Select the "Food" resource (for the sake of this example)
3. On the crowdsourcing form, click on the Food type accordion
4. Observe that the crowdsourcing form is resized and overflowing.

**Expected behavior**
The crowdsourcing form is not overflowing over other elements

**Screenshots**

<img width="838" alt="Image" src="https://github.com/user-attachments/assets/f4cf2b01-6f01-4cab-8b49-18bcc12cf7a0" />

**Additional context**
- We should set the modal with a max width that equals the toolbar width, and a max height up to the same gap between the head component and the toolbar
- We might want to create a layout grid where head component, toolbar and modal are fixed above the map but do not overflow over each other, and shrink accordingly

ravimakhijapi:
> This makes it inaccessible for the user to click/access parts that are overflowing beyond the window.

Not a bug fix but for a temporary workaround, using the shortcut for browser text zoom to zoom out should allow viewing the whole form. On macOS that's cmd+plus and cmd+minus. On Windows I think it would be ctrl instead of cmd.

PR #584, opened 2025-03-16
Adjusts width on Add Resource Forms

gcardonag:
# Pull Request

## Change Summary

This PR is primarily intended to address Design Team feedback regarding the width of the Add Resource forms.

It also includes a fix to address the Add Resource form not resetting when it is closed on mobile.

## Change Reason

Design Team feedback from a review on 03/11/2024.

## Verification [Optional]

Before
maxwidth

After
without_maxwidth

Related Issue: #575

### DIFF
--- a/src/components/AddResourceModal/ModalWrapper.jsx
+++ b/src/components/AddResourceModal/ModalWrapper.jsx
@@ -20,7 +20,7 @@ const ModalWrapper = ({ children, onExited }) => {
             left: '32px',
             bottom: '133px',
             width: '766px',
-            borderRadius: '10px',
+            borderRadius: '10px'
           }}
         >
           <Collapse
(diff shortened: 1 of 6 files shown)


src/components/AddResourceModal/ModalWrapper.jsx
```diff
@@ -44,7 +44,7 @@ const ModalWrapper = ({ children, onExited }) => {
               left: 0,
               top: 0,
               width: '100%',
-              height: '100%'
+              minHeight: '100%'
```
RNR1: What issue does minHeight 100% tries to solve? Does this element able to take more than 100% of its parent?

RNR1 (approved):
LGTM

src/components/AddResourceModal/ModalWrapper.jsx
gcardonag: That's correct, or rather the content can expand beyond 100% of the screen. When the height was set to 100%, if the form content expanded beyond the top of the screen, there is a shadow applied by MUI to the `Paper` component that would effectively look like a random line on the screen.

For context, Ron and I discussed this further on our weekly call and decided to use the `elevation` property of the Paper component to remove the shadow instead.

PR #585, opened 2025-03-18
Remove map page component and render ReactGoogleMaps directly

RNR1:
# Pull Request

## Change Summary

Render ReactGoogleMaps component directly from App.js

## Change Reason

We have a component called MapPage that only renders another component. We can render the component directly instead


gcardonag (approved):
Thanks for cleaning up a few more files 🧹

PR #586, opened 2025-03-18
Fix key issue in contributors list

RNR1:
# Pull Request

## Change Summary

Change the key prop for the contributors to account for the data migration to Supabase

## Change Reason

We changed database and not we no longer have a "Key" field for contributors, so React was complaining that the contributors list is missing the key prop


gcardonag (approved):
✅

Issue #637, opened 2025-08-19
Bug: On Chrome and Safari in Desktop, Search results are not clickable


gcardonag:
**Describe the bug**
When on phlask.me while on Google Chrome, if I click the "Search" button in the toolbar, enter a search query, and attempt to click any of the results, I am unable to click them.

**To Reproduce**
1. Go to phlask.me on Google Chrome
2. Click on the Search icon
3. Type any search query and wait for results to load
4. Attempt to click on any of the results

**Expected behavior**
Clicking on any search result after searching should redirect the map center to that selected location.

**Screenshots**
https://github.com/user-attachments/assets/ec28d42f-fad6-400d-9f41-f20b8804f12d

PR #639, opened 2025-08-20
Fix search suggestion click issue on Chrome/Safari Desktop

icycoldveins:
# Pull Request

## Change Summary

Fix search suggestion click issues on Chrome/Safari Desktop by replacing unstable crypto.randomUUID() keys with stable library keys. This prevents DOM recreation during hover state changes that was interrupting click events.

## Change Reason

Users were unable to reliably click on search suggestions in Chrome and Safari desktop browsers due to DOM elements being recreated on hover, which interrupted the click event.

## Changes
- Mobile: Changed line 143 from `crypto.randomUUID()` to `key={key}`  
- Desktop: Changed line 250 from `crypto.randomUUID()` to `key={key}`

Related Issue: #637

### DIFF
--- a/src/components/SearchBar/SearchBar.jsx
+++ b/src/components/SearchBar/SearchBar.jsx
@@ -140,7 +140,7 @@ const SearchBar = ({ search }) => {
 
                       return (
                         <div
-                          key={crypto.randomUUID()}
+                          key={key}
                           id={id}
                           role="option"
                           onMouseEnter={onMouseEnter}
@@ -247,7 +247,7 @@ const SearchBar = ({ search }) => {
                     return (
                       <div
                         className={className}
-                        key={crypto.randomUUID()}
+                        key={key}
                         id={suggestion.id}
                         role="option"
                         onMouseEnter={onMouseEnter}


icycoldveins:
# Search Suggestion Click Issue Fix

## Bug Fixed: Search Suggestion Click Issue on Chrome/Safari Desktop

**Bug**
The click issue was caused by using `crypto.randomUUID()` as React keys, which forced DOM elements to be destroyed and recreated on every render. This interrupted click event sequences when elements were destroyed between mousedown and click events.

**Context:**
- Using `crypto.randomUUID()` generates a new random key on every component render
- React sees each render as completely new elements and destroys/recreates DOM nodes
- During click sequence (mousedown -> hover state change -> re-render -> mouseup -> click), the original element gets destroyed
- The click event gets lost because the element the user started clicking no longer exists
- This created a race condition where hover-triggered re-renders broke click interactions

**Solution:**
Changed from crypto key back to the library's key

**Before:**
```jsx
const { onMouseEnter, onClick } = getSuggestionItemProps(suggestion, { className });
return (
  <div key={crypto.randomUUID()} onMouseEnter={onMouseEnter} onClick={onClick}>  // BROKEN: Random key forces DOM recreation every render
    <span>{suggestion.description}</span>
  </div>
);
```

**After:**
```jsx
const { key, onMouseEnter, onClick } = getSuggestionItemProps(suggestion, { className });
return (
  <div key={key} onMouseEnter={onMouseEnter} onClick={onClick}>  // FIXED: Stable key from library maintains DOM elements
    <span>{suggestion.description}</span>
  </div>
);
```

**Why This Works:**
- The library's `key` (from `suggestion.id`) is stable for each individual suggestion
- React maintains the same DOM element across re-renders during hover state changes
- Click event handlers remain attached to the same DOM element throughout the interaction
- Prevents the DOM destruction that was interrupting click event sequences

**Files Modified:**
- `src/components/SearchBar/SearchBar.jsx` (both mobile and desktop implementations)
  - Mobile: Line 143 - Changed from `crypto.randomUUID()` to `key={key}`
  - Desktop: Line 250 - Changed from `crypto.randomUUID()` to `key={key}`

I think this is the fix

RNR1 (approved):
Nice find! LGTM

PR #647, opened 2025-10-07
Added a progress indicator to Contact component

RaulBSanchez:
# Pull Request

## Change Summary
Added a progress indicator to the Contact Component.

## Change Reason

Users were not given any indication that the contact form was loading, adding the progress indicator allows for better user experience.

## Verification [Optional]
Screenshot 2025-10-07 at 7 35 41 PM


gcardonag (approved):
This is awesome, thanks for baking it in with such an elegantly simple solution too! It looks great on mobile too 📱

Issue #674, opened 2025-11-05
"Edit" Button Brings Up Crowdsourcing Edit Form with Resource Info Populated


marcbachan:
### Description
The "edit" button from #648 needs to bring up a dialog with the crowdsourcing edit form and inherit the informaation about the resource from which it was invoked (i.e., greyed out boxes, with the editable information available for crowdsourcing edits)

## Complexity
~ 50-100 LOC, depends on complexity of the inheritance logic

icycoldveins:
<img width="635" height="357" alt="Image" src="https://github.com/user-attachments/assets/00d229e4-d20e-4a5a-a9b6-37a38132d722" /> here is what I have right now, so I was thinking of changing the titlie to edit instead of add when we are in editing mode

marcbachan:
Per discussion - we might actually want to allow editing the address in the form. 

Note from @vontell that we also need to fetch the unique ID for the resource so that we can properly bucket suggestions by resource.

icycoldveins:
<img width="521" height="304" alt="Image" src="https://github.com/user-attachments/assets/a60762c8-d87a-4589-ab01-e439ad416ec2" />

here is what I have currently I added a TODO comment for the actual submission logic since we don't have that in place yet

marcbachan:
@icycoldveins @AnilKumar3494 Aaron is creating a new table for this to land in the database. Ideally it's a small change here to add that, but we might want to piggyback off of this to start building out views for suggested edits in the admin dashboard now that a target table exists.

gcardonag:
@icycoldveins , are you still open to working through implementing this? The original PR tied to this got closed due to the rebuild of the forms functionality. No pressure if not, we can reassign this out!

icycoldveins:
im ready boss @gcardonag 

> [@icycoldveins](https://github.com/icycoldveins) , are you still open to working through implementing this? The original PR tied to this got closed due to the rebuild of the forms functionality. No pressure if not, we can reassign this out!

Issue #648, opened 2025-10-08
Add “Edit” Button to Resource Pages


marcbachan:
## Description

Implement an “Edit” button on each resource entry that, when clicked, stores a value in its state to indicate that the user "isEditing" and a reference to the data for that particular resource entry (to help pre-fill the form with the resource's existing data)

## Complexity
~40–60 lines; simple button + event binding

PR #678, opened 2025-11-29
674: Implement edit form with resource info populated

icycoldveins:
## Change Summary

Implements the crowdsourcing edit form that pre-populates with existing resource information when users click "Suggest Edit" from the resource detail modal.

## Change Reason

Issue #674 requires the edit button (from #648) to open a crowdsourcing form with the resource's existing data pre-filled, allowing users to suggest edits to resources.

## Changes

- `src/actions/actions.js`: Added `setEditingResource` action to track the resource being edited
- `src/reducers/filterMarkers.js`: Added `editingResource` state property
- `src/components/SelectedTapDetails/SelectedTapDetails.jsx`: Replaced placeholder with meatball menu containing "Suggest Edit" and "Report" options
- `src/components/AddResourceModal/AddResourceModalV2.jsx`: 
  - Added `mapResourceToFormState()` to convert database resource format to form state
  - Auto-selects correct resource type form based on `editingResource`
  - Passes `editMode` prop to child forms
  - Defers suggestion submission workflow (TODO for admin review system, not sure how we are going to implemenet just yet.)
- `src/components/AddResourceModal/Add*.jsx` (Water, Food, Bathroom, Foraging):
  - Added `reset()` from react-hook-form to populate fields when editing
  - Dynamic title ("Edit X Resource" vs "Add X Resource")
  - Accept `editMode` prop

## Verification

The form correctly populates with resource data when clicking "Suggest Edit":
- Common fields (name, address, website, description) are pre-filled
- Resource-type-specific checkboxes reflect stored tags
- Form title changes to indicate edit mode

image

image

**Note:** Actual edit submission is deferred pending implementation of the suggestion/approval workflow.

Related Issues: #674

### DIFF
--- a/src/components/AddResourceModal/AddResourceModalV2.jsx
+++ b/src/components/AddResourceModal/AddResourceModalV2.jsx
@@ -86,71 +86,119 @@ const initialState = {
   hasFountain: false
 };
 
+const WATER_TAG_TO_FIELD = {
+  FILTERED: 'filtration',
+  BYOB: 'waterVesselNeeded',
+  ID_REQUIRED: 'idRequired',
+  WHEELCHAIR_ACCESSIBLE: 'handicapAccessible'
+};
+
+const WATER_DISPENSER_TO_FIELD = {
+  DRINKING_FOUNTAIN: 'drinkingFountain',
+  BOTTLE_FILLER: 'bottleFillerAndFountain',
+  SINK: 'sink',
+  JUG: 'waterJug',
+  SODA_MACHINE: 'sodaMachine',
+  PITCHER: 'pitcher',
+  WATER_COOLER: 'waterCooler'
+};
+
+const FOOD_TYPE_TO_FIELD = {
+  PERISHABLE: 'perishable',
+  NON_PERISHABLE: 'nonPerishable',
+  PREPARED: 'prepared'
+};
+
+const FOOD_DISTRIBUTION_TO_FIELD = {
+  EAT_ON_SITE: 'eatOnSite',
+  DELIVERY: 'delivery',
+  PICKUP: 'pickUp'
+};
+
+const FORAGE_TYPE_TO_FIELD = {
+  NUT: 'nut',
+  FRUIT: 'fruit',
+  LEAVES: 'leaves',
+  BARK: 'bark',
+  FLOWERS: 'flowers',
+  ROOT: 'root'
+};
+
+const FORAGE_TAG_TO_FIELD = {
+  MEDICINAL: 'medicinal',
+  IN_SEASON: 'inSeason',
+  COMMUNITY_GARDEN: 'communityGarden'
+};
+
+const BATHROOM_TAG_TO_FIELD = {
+  WHEELCHAIR_ACCESSIBLE: 'handicapAccessible',
+  GENDER_NEUTRAL: 'genderNeutral',
+  CHANGING_TABLE: 'changingTable',
+  SINGLE_OCCUPANCY: 'singleOccupancy',
+  FAMILY: 'familyBathroom',
+  HAS_FOUNTAIN: 'hasFountain'
+};
+
+const mapArrayToFields = (array, mapping) => {
+  if (!array) return {};
+  return array.reduce((acc, item) => {
+    const field = mapping[item];
+    return field ? { ...acc, [field]: true } : acc;
+  }, {});
+};
+
+const getStandardResourceValues = resource => ({
+  name: resource.name || '',
+  address: resource.address || '',
+  website: resource.website || '',
+  description: resource.description || '',
+  guidelines: resource.guidelines || '',
+  entryType: resource.entry_type || '',
+  latitude: resource.latitude,
+  longitude: resource.longitude,
+  isValidAddress: true,
+  pictures: [],
+  images: [],
+  handicapAccessible: false,
+  idRequired: false
+});
+
+const getWaterResourceValues = resource => ({
+  ...mapArrayToFields(resource.water?.tags, WATER_TAG_TO_FIELD),
+  ...mapArrayToFields(resource.water?.dispenser_type, WATER_DISPENSER_TO_FIELD)
+});
+
+const getFoodResourceValues = resource => ({
+  ...mapArrayToFields(resource.food?.food_type, FOOD_TYPE_TO_FIELD),
+  ...mapArrayToFields(resource.food?.distribution_type, FOOD_DISTRIBUTION_TO_FIELD),
+  organization: resource.food?.organization_name || ''
+});
+
+const getForageResourceValues = resource => ({
+  ...mapArrayToFields(resource.forage?.forage_type, FORAGE_TYPE_TO_FIELD),
+  ...mapArrayToFields(resource.forage?.tags, FORAGE_TAG_TO_FIELD)
+});
+
+const getBathroomResourceValues = resource =>
+  mapArrayToFields(resource.bathroom?.tags, BATHROOM_TAG_TO_FIELD);
+
 const mapResourceToFormState = resource => {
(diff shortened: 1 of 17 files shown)


src/components/AddResourceModal/AddWaterTap/AddWaterTap.jsx
```diff
@@ -59,8 +60,25 @@ const AddWaterTap = ({
...
+    defaultValues: {
+      name,
+      address,
+      website,
+      description
+    }
+  });
+
+  useEffect(() => {
+    reset({
```
RNR1: This shouldn’t be necessary if you have already defined the `defaultValues`

src/components/AddResourceModal/AddResourceModalV2.jsx
```diff
@@ -478,11 +574,13 @@ const AddResourceModalV2 = () => {
           checkboxChangeHandler={checkboxChangeHandler}
           textFieldChangeHandler={textFieldChangeHandler}
           isValidAddress={values.isValidAddress}
+          editMode={isEditMode}
         />
       )}
 
       {resourceForm === FORAGE_RESOURCE_TYPE && (
         <AddForaging
+          key={editingResource?.id || 'new'}
```
RNR1: What is the reason for adding the key here and in other components?

src/components/AddResourceModal/AddResourceModalV2.jsx
```diff
@@ -342,15 +423,25 @@ const AddResourceModalV2 = () => {
...
       }
 
-      addResource(newResource).then(result => {
-        dispatch(pushNewResource(newResource));
-      });
+      // If editing an existing resource, we should submit as a suggestion instead of creating a new resource
+      // TODO: Implement suggestion submission workflow with admin review/approval system
+      if (isEditMode && editingResource?.id) {
+        // For now, editing is disabled - awaiting suggestion workflow implementation
+        console.warn('Edit suggestions not yet implemented');
```
RNR1: We shouldn't commit any console logs, having a TODO comment is sufficient

src/db.js
```diff
@@ -63,4 +63,19 @@ export const getContributors = async () => {
...
+ * @param {string} resourceId The ID of the resource being suggested for edit
+ * @param {Object} suggestionData The suggested changes to the resource
+ * @returns {Promise<Object>} The submitted suggestion
+ */
+export const submitSuggestion = async (resourceId, suggestionData) => {
+    // TODO: Define and implement the full suggestion workflow:
+    // 1. Store suggestion in database
+    // 2. Create admin review/approval interface
+    // 3. Define how approved suggestions are applied to resources
+    throw new Error('Suggestion submission not yet implemented. Awaiting admin review workflow design.');
```
RNR1: Can we remove this throw statement? We should only throw errors if we need to block a specific flow. In this case, It's ok if this function will have no operation, but we should have a TODO comment for backfilling the implmentation

src/components/AddResourceModal/AddResourceModalV2.jsx
```diff
@@ -102,14 +171,26 @@ const AddResourceModalV2 = () => {
...
+    return initialState;
+  });
   const dispatch = useDispatch();
   const userLocation = useSelector(getUserLocation);
 
   const setToolbarModal = modal => {
     dispatch({ type: 'SET_TOOLBAR_MODAL', modal });
   };
 
+  useEffect(() => {
```
RNR1, line 235: Can we only use the state initialization callback to define these instead of having a useEffect hook?
We can use the `editingResource?.resource_type ?? null` as the initial value for the resource form, and the existing initialization for the values.

src/components/AddResourceModal/AddResourceModalV2.jsx
```diff
@@ -86,11 +86,80 @@ const initialState = {
   hasFountain: false
 };
 
+const mapResourceToFormState = resource => {
```
RNR1: Are we able to break this function down into multiple functions? 
- `getStandardResourceValues` that receives the resource and defaultValues (the `initialState`) and returns a new object that merges the defaultValues with the shared resource values (name, address, etc)
- for each resource: `get<ResourceType>ResourceValues` for resource-specific values. 
  - It receives the existing state and the resource and returns a new object merging the existing state with the specific values. 
  - We should iterate through the resource lists (`tags`, `dispenser_type`, etc) and modify the object if it satisfies the condition instead of using `includes` anywhere we set a value

```js
const result = resource.water.tags.reduce((prev, next) => {
  switch (tag) {
    case 'FILTERED':
      return { ...prev, filtration: true }
    ...
    default:
      return prev;
  }

}, {})
```

RNR1 (commented):
First pass

src/db.js
icycoldveins: ok gotcha

src/components/AddResourceModal/AddResourceModalV2.jsx
icycoldveins: I would have to add unmountonexit to the component. so the issue was that when the modal was closed the component stayed mounted. should I just add that to the collapse mui component? so that the component actually remounts and we can grab the correct values? not really sure what the best approach is 🤔. @RNR1

src/components/AddResourceModal/AddResourceModalV2.jsx
icycoldveins: so basically my intention was to reset all of its internal state of the component by changing the key. so when I didnt have that and switched over between different resources. The old resource form data stayed. Kinda like using the bug we had a while ago with the search function but this time intentionally to to force a fresh remount

src/components/AddResourceModal/AddResourceModalV2.jsx
icycoldveins: gotcha

src/components/AddResourceModal/AddResourceModalV2.jsx
icycoldveins: that was one of the workarounds I thought of to get around components mounting only once to show the correct data

RNR1: We should prevent switching resources while editing. I'm unsure how this is possible, but it sounds undesireable

src/components/AddResourceModal/AddResourceModalV2.jsx
RNR1: That might make sense. I think this is one of the approaches I took in my refactor PR

src/db.js
```diff
@@ -66,16 +66,15 @@ export const getContributors = async () => {
...
  * @returns {Promise<Object>} The submitted suggestion
  */
-export const submitSuggestion = async (resourceId, suggestionData) => {
-    // TODO: Define and implement the full suggestion workflow:
-    // 1. Store suggestion in database
-    // 2. Create admin review/approval interface
-    // 3. Define how approved suggestions are applied to resources
-    throw new Error('Suggestion submission not yet implemented. Awaiting admin review workflow design.');
-}
+// eslint-disable-next-line no-unused-vars
```
RNR1: nit: You can prefix with `_` to ignore the unused variables

src/components/AddResourceModal/AddResourceModalV2.jsx
icycoldveins: like if we are in edit mode and click on a different marker it wont fly to that pin?

src/components/AddResourceModal/AddResourceModalV2.jsx
icycoldveins: When I tried out unmountOnExit it only unmounted its children, and not the AddResourceModalV2 component itself. So I think the component only mounts when the app loads the first time, so im not sure how we can get around that

Issue #677, opened 2025-11-27
Fix inconsistent font styling in About and Join the Team pages


AnilKumar3494:
As a user, I would expect to see consistent font styling in the About and Join the team pages.

**Current Behavior**
Currently, there are parts of these pages that have text styling that is inconsistent with the other parts.

The "Our Story" section in About and the "Acknowledgements" section in Join the team should be updated to be consistent with the styling in the other sections.

**See the attached screenshot for an example of the inconsistency**
<img width="2145" height="1179" alt="Image" src="https://github.com/user-attachments/assets/bed54619-f3f9-4758-9367-0f1b77bf89ca" />

<img width="2057" height="1127" alt="Image" src="https://github.com/user-attachments/assets/18d605c0-704b-4ad6-b533-99761b91af80" />

PR #679, opened 2025-12-03
#677 UI Updates

AnilKumar3494:
# Pull Request

## Change Summary
Updated Text Font inconsistency in About and Join the Team pages 
Added some Review: and Fixme: s and  TODO: for possible review and updates

## Change Reason
Just UI updates, after this latest or most updated contributors can be fetched and updated.

Related Issue: #677


AnilKumar3494:
To the Contact Table TODO:
some files are taking about 2s to load .. attached Network Tab Image

--
image

src/components/Pages/Acknowledgements.tsx
```diff
@@ -1,5 +1,6 @@
 import styles from './Pages.module.scss';
 
+// REVIEW: Checked around and this component is not being used. Maybe remove?
```
RNR1: Thanks for flagging this! If you see anything like that, feel free to remove the file rather than commenting on it

src/components/Pages/Contact.tsx
```diff
@@ -26,7 +26,9 @@ const Contact = () => {
           <CircularProgress />
         </Box>
       )}
-
+      {/* //TODO: iframe is taking time to load and also airtable is loading tons of
```
RNR1: That’s a good point. Can you raise this concern in the PHLask dev channel so that we can discuss it? Putting it as a comment in the code will be less visible and a bit noisy

src/db.ts
```diff
@@ -44,6 +44,8 @@ export const addResource = {
   }
 };
 
+// FIXME: Update this db - contributorDatabaseName = 'contributors' to get Updated Contributors rendered on Join Page
```
RNR1: Same here, comments in the code gets harder to track, might be good to raise in the data slack channel that this should be updated so we can prioritize

src/components/Pages/Join.tsx
```diff
@@ -89,6 +90,8 @@ const Join = () => {
           into &quot;circles&quot; by job function:
         </p>
       </div>
+
+      {/* REVIEW: I reckon these can be dymically rednered? */}
```
RNR1: Can you elaborate more on that? And similar to other comments, if something might be a discussion, let’s use Slack instead of code comments

RNR1 (commented):
Looks good overall, just some cleanups before it's ready to go!

src/components/Pages/Contact.tsx
AnilKumar3494: Added to Slack

src/components/Pages/Join.tsx
```diff
@@ -89,38 +123,22 @@ const Join = () => {
...
-          icon={<DesignIcon />}
-          title="Design"
-          description="UX/UI Design and Resarch"
-        />
-        <Circle
-          icon={<CivicIcon />}
-          title="Civic"
-          description="Publicity and Community Engagement"
-        />
+        {ROLE_CARDS.map(card => (
```
RNR1, line 134: While this approach reduces repetition in the JSX, it makes the component's flow less intuitive, as you have to scroll up to see which circles we render. It also results in more lines of code. I'm not too concerned about mapping these circle components and would rather use this approach for dynamic data, where we don't know what data we receive, rather than for static data that doesn't change.

RNR1 (approved):
LGTM with one suggestion

PR #690, opened 2026-01-18
Persist search query

RNR1:
# Pull Request

## Change Summary

- Removed the Active Search Location Context
- Active Search Location is now stored as a query param ('q')

## Change Reason

This reduces complexity, improves performance and will be more straightforward when we'd like to rely on the active search for location based features


gcardonag (approved):
Its easier to share the map with location to others now, thank you! 🗺️

PR #692, opened 2026-01-21
feat: add ProvidedBy component for resource attribution

### LINKED ISSUE #691: Display 'Provided By' organization when viewing a resource
## Summary
When a user opens a resource on the map, display a 'Provided By' field showing which organization submitted/provided that resource.

## Use Case
Users should be able to see attribution for who contributed each resource to the map.

## Suggested Implementation
- Add a 'Provided By' field to the resource detail view
- Display the organization name that submitted the resource

icycoldveins:
## Summary
Adds "Provided By" feature to display organization attribution when viewing a resource.

- Display provider logo and/or name below "Last Modified" section
- Support multiple providers with "+X more" overflow and popover
- Logo mapping utility for local assets by provider name
- Avatar fallback (first letter) when no logo available

## Screenshot
providedby_ui

## Changes
- `src/types/ResourceEntry.ts` - Added Provider type
- `src/components/ProvidedBy/ProvidedBy.tsx` - New component with extracted ProviderAvatar and ProviderLink subcomponents
- `src/utils/providerLogos.ts` - Logo mapping utility
- `src/assets/logos/share-food-program.png` - Share Food Program logo
- `src/components/SelectedResourceDetails/SelectedResourceDetails.tsx` - Integration

## Refactoring
- Extracted `ProviderAvatar` component to handle logo/fallback avatar rendering
- Extracted `ProviderLink` component for conditional URL wrapping
- Removed redundant helper functions and inlined simple handlers
- Reduced component from 232 to 192 lines

Closes #691

Co-authored-by: Claude <claude@anthropic.com>


src/types/ResourceEntry.ts
```diff
@@ -123,6 +123,16 @@ export type ResourceStatus =
   | 'PERMANENTLY_CLOSED'
   | 'HIDDEN';
 
+/** An organization or entity that provided/contributed this resource. */
+export type Provider = {
```
RNR1, line 127: - When looking in the database, I couldn't find a provider field in the resource table. Is this feature already connected?
- Are we expecting to have more than a single provider per resource? I wonder if having a list of providers per resource makes sense, or if we should connect a single provider per resource.
- I'd expect providers to be a separate table that we join with a foreign key from the resources table when we query for the resource, so that we can define each provider and their logo, url information

icycoldveins: - nope. We'd have to add a new field for providers. 

- I should probably ask billy about the possiblity of having more than two providers for a resource.

- I was thinking of hardcoding the logos and the url just because I kinda doubt we'll have many providers, but should probably get clarification 😅.

AnilKumar3494: We should most likely add/have providers and their meta-data in our supabase and fetch from there. Will make it easier for maintaining and add any new updates later

marcbachan: @icycoldveins @RNR1 
Re: point 2, we did decide in the [planning doc](https://docs.google.com/document/d/19k5AcQ7LOlsAY0V8GBCsYUT4DdzUXTQkoOBmNEh4Q_w/edit?pli=1&tab=t.0) that we should support the possibility of multiple providers (i.e., seasonal collaboration between orgs), so a list makes sense here, even if it doesn't happen often.

Re: point 3, I agree with @AnilKumar3494 - even if we won't have many providers and there won't be a lot of turnover (if any) in the data sources, we should still maintain the database as the source of truth when/where possible. For example, if we have an entities table for providers, we can just add a text column to store the URL for the logo, site, etc. In Share Food's case, I know their logo is [here](https://www.sharefoodprogram.org/wp-content/themes/sharefood-theme/images/svg/share-food-program-logo.svg) from poking around their site.

src/components/ProvidedBy/ProvidedBy.tsx
```diff
@@ -0,0 +1,214 @@
...
+                            underline="hover"
+                            color="inherit"
+                          >
+                            {provider.name}
+                          </Link>
+                        ) : (
+                          provider.name
+                        )
+                      }
+                      primaryTypographyProps={{ fontSize: 14 }}
```
RNR1: This prop has been deprecated, Use `slotProps.primary` instead
https://mui.com/material-ui/api/list-item-text/#list-item-text-prop-primaryTypographyProps

src/components/ProvidedBy/ProvidedBy.tsx
```diff
@@ -0,0 +1,214 @@
...
+                horizontal: 'left'
+              }}
+            >
+              <List dense sx={{ minWidth: 200 }}>
+                {hiddenProviders.map((provider, index) => {
+                  const logoUrl = provider.logo_url || getProviderLogo(provider.name);
+                  return (
+                  <ListItem key={`${provider.name}-${index}`}>
+                    <ListItemAvatar sx={{ minWidth: 48 }}>
+                      {shouldShowImage(logoUrl) ? (
```
RNR1: nit: The usage of box wrapping an img element is a bit non-standard, but it might make more sense as an implementation detail for an `ImageWithInitialsAvatarFallback` component

src/components/ProvidedBy/ProvidedBy.tsx
```diff
@@ -0,0 +1,214 @@
...
+  ListItemText,
+  Link,
+  Box
+} from '@mui/material';
+import InfoOutlinedIcon from '@mui/icons-material/InfoOutlined';
+import { type Provider } from 'types/ResourceEntry';
+import { getProviderLogo } from 'utils/providerLogos';
+
+type ProvidedByProps = {
+  providers: Provider[] | null | undefined;
```
RNR1: Since it's desirable to have multiple providers, I'd suggest this to be an optional prop that only accept `Provider[]` and fallbacks to an empty array, so that we don't need to null check it.

marcbachan:
@vontell FYI on the discussion above:
- We need to add a `provider` column (or table, given that we need to support multiple providers per resource where needed) for resources
- We also need to either store the image for the provider logo in the database, or a URL to the publicly available resource

src/assets/logos/share-food-program.png
RNR1, line 1: I'd like to avoid having static assets of other organization logos in the codebase - let's ensure we only pull these from the database

src/components/ProvidedBy/ProvidedBy.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+  const handleImageError = (logoUrl: string) => {
+    setFailedImages(prev => new Set(prev).add(logoUrl));
+  };
+
+  const shouldShowImage = (logoUrl: string | undefined) => {
+    return logoUrl && !failedImages.has(logoUrl);
+  };
+
+  const renderProviderItem = (provider: Provider) => {
```
RNR1: This might be nice to have as its own component rather than part of this PR

src/utils/providerLogos.ts
RNR1, line 1: since we're not having assets in the code, we shouldn't have this map in the code too

src/components/ProvidedBy/ProvidedBy.tsx
```diff
@@ -0,0 +1,214 @@
...
+          <InfoOutlinedIcon
+            sx={{ fontSize: 16, color: '#60718C', cursor: 'help' }}
+          />
+        </Tooltip>
+      </Stack>
+
+      <Stack gap={1}>
+        {visibleProviders.map((provider, index) => (
+          <div key={`${provider.name}-${index}`}>
+            {renderProviderItem(provider)}
```
RNR1: once this is a separate component, call this as a JSX element instead of a function

src/components/SelectedResourceDetails/SelectedResourceDetails.tsx
```diff
@@ -208,14 +209,17 @@ const SelectedResourceDetails = ({
...
+          <Stack gap={1.5}>
+            {!isMobile && resource.last_modified ? (
+              <Stack>
+                <Typography fontWeight={600}>Last Modified</Typography>
+                <Typography color="#60718C">
+                  {new Date(resource.last_modified).toDateString()}
+                </Typography>
+              </Stack>
+            ) : null}
+            <ProvidedBy providers={resource.providers} />
```
gcardonag: In order to make this work with the data structure we set up this week in Supabase, we'll look into making the following changes to this branch next week (if you're around to join the fun @icycoldveins !)
- Update this line to reference a `providers` variable
- Add a `const providers` variable as this component is initializing that references a new database-querying export in `src/services/db.ts`.
- Add a database query that does a join across the resource and providers Tables using a many-to-many join and return a list of Provider-type objects: https://supabase.com/docs/guides/database/joins-and-nesting#many-to-many-joins

Then we can give this a test with the test datapoint we added to the beta site 🤞

src/components/forms/FormAddressField/FormResourceAddressField.tsx
```diff
@@ -114,12 +114,15 @@ const FormResourceAddressField = ({
       renderInput={({ inputProps, ...params }) => (
         <TextField
           {...params}
-          inputProps={{
```
RNR1: Note that this has been left as-is due to compatibility issues with the Autocomplete component. When I tried migrating it, it broke. If you're changing this, make sure to test that it works the same way, and there are no regressions

src/components/SelectedResourceDetails/SelectedResourceDetails.tsx
icycoldveins: im ready sounds good! @gcardonag

gcardonag (approved):
🍏

PR #693, opened 2026-01-21
ak/update-contributors

AnilKumar3494:
Current and Past Contributors fetching and rendering updated

Same as: https://github.com/phlask/phlask-map/pull/689


src/components/ContributorsList/ContributorsListItem.tsx
```diff
@@ -28,8 +18,7 @@ const ContributorsListItem = ({ contributor }: ContributorsListItemProps) => (
       </Box>
       <Stack direction="row">
         <Typography component="span" variant="subtitle2">
-          {contributorToCircleName(contributor)}
-          {contributor.convener ? ', Convener' : ''}
```
RNR1, line 32: Add the convener back here

src/hooks/queries/useContributorsQuery.ts
```diff
@@ -15,7 +15,7 @@ const useGetContributorsQuery = () =>
         (prev, next) => {
           if (next.is_active) {
             prev.current.push(next);
-          } else {
+          } else if (next.is_contributor === 'checked') {
```
RNR1: Once the `is_contributor` is changed to a boolean field, let's make this a filter when we query Supabase instead of doing this filtering at the frontend, so that Supabase only returns contributors in the first place

src/types/Contributor.ts
```diff
@@ -1,16 +1,13 @@
 export type Contributor = {
-  civic_member: boolean;
-  convener: boolean;
```
RNR1, line 3: make sure to add `is_convener` back once it's populated

AnilKumar3494 (commented):
should be gtg

RNR1 (approved):

PR #694, opened 2026-01-21
optimized location features

RNR1:
# Pull Request

## Change Summary

- Location prompting no longer happens on the app load
- Changed the Main button from "Near me" to "Nearby" to reflect a more semantic lookup by sites nearby your search location or your current location
- The last active search is now persisted
- If a user interacts with a location-based feature* - we:
  - prompt them to enable location services
    - If enabled:
      - location-based features will use the user's location
      - We will display and persist this location on the search bar
    - If not enabled: 
      - location-based features will fallback to the user's active search
      - If none exists, a toast alert will be displayed to provide feedback to the user

* Location-based features can be: Nearby button, Estimated walk time, use my location in crowdsourcing forms

## Change Reason

Prompting the user for location before interacting with the app is a violation. We should only ask for the user location if they interact with a location-based feature


src/hooks/queries/useWalkingDurationQuery.ts
AnilKumar3494, line 1: this feature is lit 🔥

vontell (approved):

PR #703, opened 2026-02-12
Remove unnecessary use effect from the codebase

RNR1:
# Pull Request

## Change Summary

Remove the `useOnClickOutside` hook that we mount on the choose resource modal

## Change Reason

It's redundant, as this behavior is already handled by the Modal component


gcardonag (approved):
🧹

Issue #686, opened 2026-01-14
Swap the contact form from airtable to a form that uploads to supabase


vontell:
Swap the contact form from airtable to a form that uploads to supabase

Issue #680, opened 2025-12-04
Fix: Contact Form Loading Time


AnilKumar3494:
**Problem:**
Airtable iframe (form to contact) is taking about 2s to 3s to load in the 'Contact' section of the page.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to https://phlask.me/
2. Click on 'Hamburger Menu >> Contact'
3. Form Loading Time

**Expected behavior**
Loading the form quickly <1s

**Screenshots**
<img width="720" height="400" alt="Image" src="https://github.com/user-attachments/assets/16c3cabd-82b4-427f-afff-d15ca1ad37ee" />

**Potential Fixes**
Custom Form and API proxy to Airtable and maybe adding loading skeleton for UX.

PR #710, opened 2026-02-16
update feedback form

AnilKumar3494:
## Change Summary

This updates the feedback/contact form page. Instead of loading AirTable iframe that was taking about 4 to 5 sec every time, this replaces it with a custom form and does insert to Supabase 'airtable_feedback' table on submit.

## Change Reason
Makes the form load and submission instantly and gives full control to us.

This also resolves the below:
[#680](https://github.com/phlask/phlask-map/issues/680)
[#686](https://github.com/phlask/phlask-map/issues/686)
[#687](https://github.com/phlask/phlask-map/pull/687)


src/components/Pages/Contact.tsx
RNR1, line 1: Use zod for schema validation, and react-hook-form for the form management. Use the existing form fields that are already wrapped with React Hook Form. You should check the existing crowdsourcing forms as reference

RNR1 (changes requested):

src/components/Pages/Contact.tsx
```diff
@@ -1,42 +1,191 @@
...
+  },
+  '& .MuiInputLabel-root.Mui-focused': {
+    color: PRIMARY_COLOR
+  }
+};
+const SCHEMA = feedbackFormSchema;
 
 const Contact = () => {
-  const [loading, setLoading] = useState(true);
+  const [status, setStatus] = useState<'idle' | 'success' | 'error'>('idle');
```
RNR1: For API calls - we should use TanStack Query. In this case, you should create a hook that consumes `useMutation`. See existing forms as a reference

src/components/AddResourceModal/AddBathroomForm/AddBathroomForm.tsx
```diff
@@ -94,7 +94,7 @@ const AddBathroomForm = ({ onGoBack, onComplete }: AddBathroomFormProps) => {
                 name="bathroom.tags"
                 label="Helpful info"
                 options={tagOptions}
-                labelPlacement="start"
+                labelPlacement="end"
```
RNR1: Can we create the new form without affecting existing forms? Is this change necessary?

src/components/forms/FormCheckBoxField/FormCheckBoxField.tsx
```diff
@@ -0,0 +1,59 @@
...
+  sx?: TextFieldProps['sx'];
+};
+
+const FormCheckboxField = <Values extends FieldValues>({
+  name,
+  label,
+  labelPlacement = 'end',
+  sx
+}: FormCheckboxFieldProps<Values>) => {
+  const { control } = useFormContext<Values>();
```
RNR1: Ideally, we only use controlled fields when we have no other choice (e.g., for non-standard form fields that doesn't support native HTML inputs). using uncontrolled form fields (using the `register` method) should be the preferred way, as it will not trigger rerender for each change. See the existing implementation as a reference

src/components/forms/FormCheckBoxField/FormCheckBoxField.tsx
```diff
@@ -0,0 +1,59 @@
...
+  type Path
+} from 'react-hook-form';
+import { type ReactNode } from 'react';
+
+type FormCheckboxFieldProps<Values extends FieldValues> = {
+  name: Path<Values>;
+  label: ReactNode;
+  helperText?: string;
+  labelPlacement?: 'bottom' | 'end' | 'start' | 'top';
+  sx?: TextFieldProps['sx'];
```
RNR1: For consistency, we should avoid exposing style changes for generic components. This will open the door to inconsistent styling across pages

src/components/forms/FormCheckboxListField/FormCheckboxListField.tsx
```diff
@@ -44,7 +44,6 @@ const FormCheckboxListField = <Values extends FieldValues>({
             label={option.label}
             labelPlacement={labelPlacement}
             value={option.value}
-            sx={{ justifyContent: 'space-between' }}
```
RNR1, line 47: What is the reason for this change?

src/components/forms/FormTextField/FormTextField.tsx
```diff
@@ -10,23 +10,29 @@ import {
...
   label: ReactNode;
+  placeholder?: string;
   disabled?: boolean;
   helperText?: ReactNode;
   fullWidth?: boolean;
   required?: boolean;
   multiline?: boolean;
   minRows?: number;
+  slotProps?: TextFieldProps['slotProps'];
+  sx?: TextFieldProps['sx'];
```
RNR1, line 20: Same here

src/components/Pages/Contact.tsx
```diff
@@ -1,42 +1,172 @@
...
+              name="feedback"
+              label="Feedback"
+              placeholder="Share your feedback and thoughts"
+              multiline
+              minRows={3}
+              fullWidth
+              sx={textFieldFocus}
+              helperText="Please do not include any sensitive personal information."
+              slotProps={{
+                formHelperText: {
```
RNR1: What is the reason for providing different styles for this particular helper text? We should have consistent styling in all places.

src/components/Pages/Contact.tsx
```diff
@@ -1,42 +1,172 @@
...
+import ClearFormIcon from 'icons/ClearFormIcon';
+import FormWrapper from 'components/forms/FormWrapper/FormWrapper';
+import FormTextField from 'components/forms/FormTextField/FormTextField';
+import FormCheckboxField from 'components/forms/FormCheckBoxField/FormCheckBoxField';
+
+type FormValues = FeedbackFormValues;
+
+const TITLE = 'Contact Us';
+const PRIMARY_COLOR = '#10b6ff';
+const textFieldFocus = {
```
RNR1: We should use the existing styles and avoid overriding in specific screens

src/components/Page/Page.tsx
```diff
@@ -13,6 +13,7 @@ const Page = ({ children, title, 'data-cy': dataCy }: PageProps) => {
       sx={{
         color: '#60718c',
         maxHeight: '55vh',
+        width: '100%',
```
RNR1: Usually, 100% width will be the default. Is that not the case here?

src/components/forms/FormWrapper/FormWrapper.tsx
RNR1, line 1: I don't see a reason to wrap the form element with a box - I think you should be good using `<form onSubmit={handleSubmit(onSubmit))>` inline instead

src/components/Pages/Contact.tsx
```diff
@@ -1,42 +1,172 @@
...
+      interest: false
+    },
+    resolver: zodResolver(SCHEMA)
+  });
+
+  const { reset } = methods;
+
+  const onSubmit = (feedbackData: FormValues) => {
+    addFeedbackMutate(feedbackData, {
+      onSuccess: () => {
```
RNR1: The success state should replace the entire screen, so I think you can create a component that conditionally renders the success screen, similar to how we currently do with crowdsourcing forms. In that case, I don't think you have to reset the form, just add an `if (isSuccess)` before this component return.

Image

Image

src/components/Pages/Contact.tsx
```diff
@@ -1,42 +1,172 @@
...
-            height: '100%',
-            display: 'flex',
-            alignItems: 'center',
-            justifyContent: 'center'
-          }}
+    <Page title={TITLE} data-cy="contact-us">
+      <FormProvider {...methods}>
+        <FormWrapper
+          onSubmit={onSubmit}
+          sx={{ width: { xs: '90%', md: '75%' } }}
```
RNR1: If you need to make stylistic changes to the form - you should apply them on a wrapper component instead of the form element directly

src/components/Pages/Contact.tsx
```diff
@@ -1,42 +1,172 @@
...
+                  backgroundColor: PRIMARY_COLOR,
+                  '&:hover': {
+                    backgroundColor: PRIMARY_COLOR,
+                    boxShadow: `0 0 10px ${PRIMARY_COLOR}80`
+                  }
+                }}
+              >
+                {isPending
+                  ? 'Submitting...'
+                  : isSuccess
```
RNR1: Avoid nested ternaries

src/components/Pages/Contact.tsx
```diff
@@ -1,42 +1,172 @@
...
+              name="interest"
+              label="I'm interested in helping PHLASK with future research"
+              labelPlacement="end"
+              sx={{
+                '&.Mui-checked': { color: PRIMARY_COLOR }
+              }}
+            />
+
+            <Stack direction="row" gap={5} mt={2}>
+              <Button
```
RNR1: The design doesn't show a reset button for this form, so I don't think we have to show one

src/components/Pages/Contact.tsx
```diff
@@ -1,42 +1,172 @@
...
+                onClick={handleClear}
+                sx={{ color: PRIMARY_COLOR }}
+                startIcon={<ClearFormIcon />}
+              >
+                clear form
+              </Button>
+              <Button
+                variant="contained"
+                type="submit"
+                disabled={isPending || isSuccess}
```
RNR1: you can use the Button `loading` prop for that

src/icons/JoinDataIcon.tsx
RNR1, line 1: What is the reason for this icon change?

src/services/db.ts
```diff
@@ -11,6 +11,7 @@ const databaseApiKey =
   import.meta.env.VITE_DB_API_KEY ||
   'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6IndhbnR5Y2Zibnp6b2NzYnRocXpzIiwicm9sZSI6ImFub24iLCJpYXQiOjE3MzcwNDY2OTgsImV4cCI6MjA1MjYyMjY5OH0.yczsMOx3Y-zsWu-GjYEajIb0yw9fYWEIUglmmfM1zCY';
 const contributorDatabaseName = 'airtable_contributors';
+const feedbackDatabaseName = 'airtable_feedback';
```
RNR1: Is this table sourced from Airtable? Otherwise - I think a better name for it would be user_feedbacks

src/types/ResourceEntry.ts
```diff
@@ -178,3 +178,10 @@ export type ResourceEntry = {
   /**  If the resource_type is BATHROOM, the information about the bathroom resource. */
   bathroom?: BathroomInfo | null;
 };
+
+export type FeedbackForm = {
```
RNR1: The ResourceEntry typing file is dedicated for Resource related type definitions. Since this is a feedback form, it should live in a seperate file

src/components/forms/FormCheckboxListField/FormCheckboxListField.tsx
AnilKumar3494: This makes the labelPlacement to use 'end' instead of 'start' (the labels are rendered to the right of the checkbox) and would also not effect the spacing if CheckboxList component were to be reused again anywhere else.

src/components/AddResourceModal/AddBathroomForm/AddBathroomForm.tsx
AnilKumar3494: This change has the same visuals as before. 

Since the labels are rendered to the right of the checkbox having labelPlacement=end seemed initiative.

src/components/forms/FormCheckboxListField/FormCheckboxListField.tsx
AnilKumar3494: **Helpful Info**


**Current:**
Image

After Change: t

Image

src/components/forms/FormTextField/FormTextField.tsx
AnilKumar3494: removed slotProps but need sx to have width as per design

src/icons/JoinDataIcon.tsx
AnilKumar3494: With New Icon Change:
Image

Current version of the app:
Is using the same icon from Data and Civic

Image

src/services/db.ts
AnilKumar3494: This is sourced from Airtable for now; however, once these changes are live, we won't need to use Airtable anymore ... just Supabase.

Guess user_feedbacks will make more sense now

src/icons/index.ts
```diff
@@ -43,5 +43,4 @@ export { default as UsersIcon } from './UsersIcon';
 export { default as WaterIcon } from './WaterIcon';
 export { default as WaterIconChooseResource } from './WaterIconChooseResource';
 export { default as WaterIconV2 } from './WaterIconV2';
-export { default as PhlaskContactSubmit } from './PhlaskContactSubmit';
```
AnilKumar3494, line 46: This is not being used anywhere and also this SVG was super heavy and was breaking the app when used. so removed

src/contexts/HeaderContext.tsx
```diff
@@ -1,5 +1,5 @@
 import About from 'components/About/About';
-import Contact from 'components/Pages/Contact';
```
AnilKumar3494, line 2: other pages are not in the Pages folder so made Contact to match the pattern

src/components/Page/Page.tsx
```diff
@@ -12,8 +12,9 @@ const Page = ({ children, title, 'data-cy': dataCy }: PageProps) => {
     <Stack
       sx={{
         color: '#60718c',
-        maxHeight: '55vh',
```
AnilKumar3494, line 15: this height of 85vh for mobile to reflect figma design

AnilKumar3494 (commented):
she's working

src/components/SocialLinks/SocialLinks.tsx
```diff
@@ -0,0 +1,44 @@
+import { Link, Stack, Typography } from '@mui/material';
+import { SocialFacebook, SocialInstagram, SocialLinkedin } from 'icons';
+
+function SocialContacts() {
```
RNR1: For consistency with the rest of our components - this should be an arrow function
```tsx
const SocialContacts = () = > (
  ...
)
```

src/components/SocialLinks/SocialLinks.tsx
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <Stack direction={'column'} alignItems={'center'} gap={1}>
```
RNR1: For consistency, we shouldn't wrap string values with curly braces in JSX

src/components/SocialLinks/SocialLinks.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <Typography variant="h6">Follow PHLASK</Typography>
+      <Stack direction={'row'} gap={4}>
```
RNR1, line 8: same here

src/components/forms/FormCheckboxListField/FormCheckboxListField.tsx
RNR1: Image

The way this checkbox list field is currently implemented aligns with the designs for this
 feature, so if we change it, it would drift away from the design intention. Can we not change this unless there's a reference in the design where this is incorrect?

src/components/Page/Page.tsx
```diff
@@ -8,12 +9,14 @@ type PageProps = {
 };
 
 const Page = ({ children, title, 'data-cy': dataCy }: PageProps) => {
+  const isMobile = useIsMobile();
```
RNR1: nit: if you only use the hook for a style change, you should use the sx function override:

```tsx
sx={(theme) => ({
  maxHeight: '85vh',
  [theme.breakpoints.up('md')]: {
    maxHeight: '55vh',
  },
})
```

src/components/forms/FeedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  const handleFormSubmit: SubmitHandler<FormValues> = feedbackData => {
+    onSubmit(feedbackData);
+  };
+
+  return (
+    <FormProvider {...methods}>
+      <Typography variant="h6">Share Feedback</Typography>
+      <form onSubmit={methods.handleSubmit(handleFormSubmit)}>
```
RNR1: nit: destructure the relevant methods from the methods object before the return statement

src/components/forms/FeedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+          <FormTextField<FormValues>
+            name="feedback"
+            label="Feedback"
+            placeholder="Share your feedback and thoughts"
+            multiline
+            minRows={3}
+            required
+            helperText="Please do not include any sensitive personal information."
+            sx={{ width: isMobile ? '95%' : '800px' }}
```
RNR1: I'd like to avoid fixed widths. Should we use the `fullWidth` prop instead, then define a maxWidth on the parent Stack component?

src/components/forms/FeedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,100 @@
+import { FormProvider, useForm, type SubmitHandler } from 'react-hook-form';
+import feedbackFormSchema, {
+  type FeedbackFormValues
+} from 'schemas/feedbackFormSchema';
+import FormTextField from '../FormTextField/FormTextField';
+import { Stack, Button, Typography } from '@mui/material';
+import FormCheckboxField from '../FormCheckBoxField/FormCheckBoxField';
```
RNR1: use an absolute import here

src/components/EmailConnect/EmailConnect.tsx
```diff
@@ -0,0 +1,19 @@
+import { Link, Stack, Typography } from '@mui/material';
+import useIsMobile from 'hooks/useIsMobile';
+
+function Connect() {
+  const isMobile = useIsMobile();
+  return (
+    <Stack direction={'column'} gap={1}>
```
RNR1: avoid wrapping string prop values with curly braces

src/components/EmailConnect/EmailConnect.tsx
```diff
@@ -0,0 +1,19 @@
+import { Link, Stack, Typography } from '@mui/material';
+import useIsMobile from 'hooks/useIsMobile';
+
+function Connect() {
```
RNR1: use arrow function for consistency

src/components/AddResourceModal/AddWaterForm/AddWaterForm.tsx
```diff
@@ -107,7 +107,7 @@ const AddWaterForm = ({ onGoBack, onComplete }: AddWaterFormProps) => {
                 name="water.tags"
                 label="Helpful info"
                 options={tagOptions}
-                labelPlacement="start"
+                labelPlacement="end"
```
RNR1: similar to the other comment - this aligns with the design mock ups - we shouldn't touch it

src/components/forms/FeedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+  };
+
+  return (
+    <FormProvider {...methods}>
+      <Typography variant="h6">Share Feedback</Typography>
+      <form onSubmit={handleSubmit(handleFormSubmit)}>
+        <Stack
+          gap={2}
+          sx={theme => ({
+            mb: 4,
```
RNR1: avoid abbreviated properties

src/components/forms/FeedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+const SCHEMA = feedbackFormSchema;
+
+const FeedbackForm = ({ onSubmit, isPending }: FeedbackFormProps) => {
+  const methods = useForm<FormValues>({
+    defaultValues: {
+      name: '',
+      email: '',
+      feedback: '',
+      interest: false
```
RNR1: rename the table field to `is_interested_in_contributing`

src/components/forms/FeedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+          />
+
+          <Button
+            type="submit"
+            variant="contained"
+            loading={isPending}
+            sx={theme => ({
+              backgroundColor: '#10B6FF',
+              borderRadius: '8px',
+              width: '168px',
```
RNR1: remove fixed with

src/components/forms/FormCheckboxField/FormCheckboxField.tsx
```diff
@@ -0,0 +1,30 @@
...
+  label,
+  labelPlacement = 'end'
+}: FormCheckboxFieldProps<Values>) => {
+  const { register } = useFormContext<Values>();
+  const field = register(name);
+
+  return (
+    <FormControl>
+      <FormControlLabel
+        {...field}
```
RNR1: register here instead of in a variable

src/schemas/feedbackFormSchema.ts
```diff
@@ -0,0 +1,11 @@
+import * as z from 'zod';
+
+const feedbackFormSchema = z.object({
+  name: z.string().min(1, 'Name is required'),
+  email: z.email().min(1, 'Email is required'),
+  feedback: z.string().min(1),
+  interest: z.boolean()
```
RNR1: update field name

src/types/FeedbackEntry.ts
```diff
@@ -0,0 +1,6 @@
+export type FeedbackForm = {
+  name: string | null;
```
RNR1: remove nulls for all fields

AnilKumar3494 (commented):
🚀

RNR1 (approved):
LGTM

PR #718, opened 2026-02-25
change nearby behavior

RNR1:
# Pull Request

## Change Summary

When the user clicks the "Nearby" button - it would remember either their location (if granted permission) or the active search location. We have changed this to remember the closest location instead

## Change Reason

It felt like a better UX once the user refreshed the page

### DIFF
--- a/src/components/Toolbar/Toolbar.tsx
+++ b/src/components/Toolbar/Toolbar.tsx
@@ -60,8 +60,8 @@ const Toolbar = () => {
       setSelectedResource(closest, prev);
       onChangeActiveSearchLocation(
         {
-          lat: location.lat,
-          lng: location.lng,
+          lat: closest.latitude,
+          lng: closest.longitude,
           placeId: closest.gp_id || undefined
         },
         prev


gcardonag (approved):
🗺️

src/components/Toolbar/Toolbar.tsx
```diff
@@ -60,8 +60,8 @@ const Toolbar = () => {
       setSelectedResource(closest, prev);
       onChangeActiveSearchLocation(
         {
-          lat: location.lat,
-          lng: location.lng,
+          lat: closest.latitude,
```
AnilKumar3494, line 63: ## ** When location services turned on**

**case1 - situation: When clicked on a pin (Dilworth Park)**
Est walk time is 3345min in this example from my 'current location' which is perfect

Image

--
**case2 - situation: I search for a location (Drexel Uni in this example) and then click on a pin (Dilworth Park)**
Est walk time: is still 3345min which is still being calculated from my current location.

_**Expected: Should show est walk time from active seach location?**_

Image

--
**case3 - situation: I click nearby button**
It is taking me to Firefighter's Field with est walk time 1483min which i reckon is the closest one since I'm in west haven. This is working perfectly!

Image

--
**case4 - situation: i search for location (Drexel University) in this case and click nearby** 
It is taking me back to Firefighter's Field which is based on my current location and not active seach location. And in the process the search bar is getting updated to the resource address 

**_I think general intuition would be get a resource close to active search (Drexel in this case) as that would be the users motive._**

Video Link: [https://drive.google.com/file/d/1AWU2HUfvkDlnC5bCR-GEl30qMWR2GFmX/view?usp=sharing](https://drive.google.com/file/d/1AWU2HUfvkDlnC5bCR-GEl30qMWR2GFmX/view?usp=sharing)

Image


## ** When Location services turned off**
**Case1-situation: i click on a pin**
It shows: Est. walking time: Not Available ... this is perfect as there is not location provided

Image

--
**Case2-situation: i click Nearby** 
It says need to enable location services alert. This isperfect as well

Image

--
**Case3: situation: I search for a location (Drexel) and click on a pin (Dilworth Park)**
It shows est walk time of 29m from current search this is perfect as well

Image

--
**Case4: situation: I search for a location (Drexel) and click the nearby** 
It is taking me to the closest resource available i.e (Starbucks 3141 Chestnut St) which is correct. However, in this process the search bar is changing to 'Starbucks' and the Est. walking time is rendering as 'Not Available'.
**_Expected: Should show est walk time from Active Searched Location (Drexel) to closest resource (Starbucks 3141 Chestnut St)_**

Video Link: [https://drive.google.com/file/d/1XoTq0DVnObWe3ZFaS5RcpQe1z1xEJvOD/view?usp=sharing](https://drive.google.com/file/d/1XoTq0DVnObWe3ZFaS5RcpQe1z1xEJvOD/view?usp=sharing)

Image

gcardonag: @RNR1 , just to make sure we don't lose our train of thought on this one. The only scenario that we believe may need deeper review of is Case4 when Location Sharing is not enabled. At a glance, it seems like showing walking times should be consistent with the Nearby feature, where both act relative to the latest Search (when location sharing isn't available)

PR #720, opened 2026-03-10
697 - Data Sourcing for Sharing Excess

### LINKED ISSUE #697: Data sourcing for Sharing Excess food distribution
No; we would like a new feature that can pull food distribution information from Sharing Excess' website

Pull food distribution data from Sharing Excess and display it as Food sites on the app

None

**Additional context**
This feature is analogous to pulling data from Share Food Program, but will require different technical development

-------------

We would like to pull data from the [Sharing Excess website embedded Google Calendar](https://www.sharingexcess.com/find-food). Every distribution listed on their site is a **one-time pop-up**. The current modal design cannot render 'temporary taps,' so we will need to have some scripted code that makes the site inactive (or archived) after the scheduled distribution is over.

The data that can be scraped include:

(PHLASK database) : (corresponding Google event field)
- Name : Event Name
- Location : Event location
- Description : Event date
- Description : Event time
- Description : Event description

This means that three data fields from the event information will all need to be condensed into the singular Description field in a site modal

As a rule, the site should:
- appear 1 week prior to the event date
- disappear (be made inactive) at the date/time of event conclusion

This will help prevent too many temporary taps from appearing on the map

marcbachan:
# Pull Request

## Change Summary
FYI: Claude helped quite a bit here in building out a basic CLI component and adding Supabase helper functions. Extra scrutiny on those is welcome.

Addresses #697. Introduces a standalone Python script that is designed to pull down events from a public Google Calendar, such as Sharing Excess, and normalize the retrieved events to be able to store them in the `resources` table in Supabase.

As the `resources` table does not have start/end date fields, these are pulled from the site and inserted into the description with some clear delimiters, like:

```
[[ start: 2026-03-10T15:00:00-04:00 | end: 2026-03-10T17:00:00-04:00 ]]
```

This allows us to do some post-processing/filtering to determine whether the event is "live" or not. 

We can do this scrape periodically by using the `LOOK_FORWARD_DAYS` property to get all events for a specific window into the future, or just do this monthly in one of the PHLASK sessions or something. Not sure how we want to handle.

## Change Reason

Billy summed this up quite nicely on #697. Essentially, we would like to be able to actively maintain "live" food sites posted by Sharing Excess and help them and us get the word out a little easier.

## Verification [Optional]

Here is an example of a CSV debug output that we can get by using the basic CLI component that Claude helped write:

```shell
 python calendar_to_supabase.py --csv 
```
[events.csv](https://github.com/user-attachments/files/25880260/events.csv)

These records can then be written to the DB either directly with CSV import in Supabase, or enter the credentials in the `.env` file here and run the script with the helper to write them to the `resources` table. 

Related Issue: #697


marcbachan:
@icycoldveins Got Claude's help with the edge function (`supabase/functions/sync-sharing-excess/index.ts`) for this in a new directory for Supabase sync operations. Let me know if this lines up with what you had tried out, and let me know if you have any suggestions. 

@gcardonag @vontell curious to get your thoughts on this regarding automating and maintaining the sync script by using edge functions. Otherwise we can just use the Python script with Lambda in AWS like the other one.

data/scrape-sharing-excess-calendar/calendar_to_supabase.py
```diff
@@ -0,0 +1,386 @@
...
+    if args.no_hours:
+        resources = [{**r, "hours": None} for r in resources]
+        print("Hours field omitted (--no-hours).")
+
+    if args.csv:
+        save_csv(resources, args.csv)
+    else:
+        supabase = get_supabase_client()
+        delete_by_creator(supabase)
+        insert_resources(supabase, resources)
```
RRodriguez26: Having a discussion with Ron and Añil, what is the use of the delete function? I also heard that we want to delete anything overall, is there a reason why might we use the delete functionality in the database rather than just update?

RRodriguez26:
I also heard that these data scripts should be in its own repo, we see that there is a [repo](https://github.com/phlask/data-circle) dedicated to it but we are not sure if this is the right one.

marcbachan:
> I also heard that these data scripts should be in its own repo, we see that there is a [repo](https://github.com/phlask/data-circle) dedicated to it but we are not sure if this is the right one.

Yep, I'm going to put up a joint PR for this script and the other one on that repo. I don't think anyone can recall if there was another reason for it, so it's a good fit.

data/scrape-sharing-excess-calendar/calendar_to_supabase.py
marcbachan: @RRodriguez26 Updating is definitely better, and I can tweak this a bit further to do it properly. I settled on delete so that we could update our database with current data, which was becoming a problem. Now that it's updated, I can revisit this and implement it more intelligently. 

One of the key issues with doing the update route was that recurring events, despite having many distinct occurrences, all collide under one `gp_id`, so the script needs to handle this on repeated syncs and update the timestamps for whatever occurrence of that event is up next instead of processing every occurrence as a unique resource. 

But overall, yes, you all are right. Deletion creates an issue with churned resource IDs, especially if there are crowdsourced edits using that ID as a foreign key. In the long term that's not sustainable, so I'll work on resolving that update issue for the recurring events.

PR #755, opened 2026-06-02
Feature/provided by v2

AnilKumar3494:
ALT txt update


gcardonag (approved):
🖌️ Thank you for taking the live feedback today and including it here!

PR #766, opened 2026-08-12
Fixes/suggest edit forms

AnilKumar3494:
# Pull Request

## Change Summary

- Address prefill for all Suggest Edits
- Close-X --- fixes: close an edit midway, then click Add Site → you'd get the old edit form instead of a blank Add form.
- Remount per resource (key)  --- fixes: open Suggest Edit on resource B right after A → B showed A's data.
- The Foraging and Food resource 'Suggest edits' were crashing without a catch, updated the schema and added defaults.

Related: #674 #678 #753

### DIFF
--- a/src/components/BathroomResourceForm/BathroomResourceForm.tsx
+++ b/src/components/BathroomResourceForm/BathroomResourceForm.tsx
@@ -18,6 +18,7 @@ type BathroomResourceFormProps = {
   defaultValues?: ResourceEntry | null;
   onSubmit: (values: FormValues) => void;
   onGoBack?: VoidFunction;
+  onClose?: VoidFunction;
   isSubmitting?: boolean;
 };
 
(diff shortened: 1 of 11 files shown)


src/components/BathroomResourceForm/BathroomResourceForm.tsx
```diff
@@ -30,13 +31,15 @@ const BathroomResourceForm = ({
...
-  onGoBack
+  onGoBack,
+  onClose: onCloseProp
 }: BathroomResourceFormProps) => {
   const { setToolbarModal } = useToolbarContext();
   const onClose = () => {
     if (onGoBack) {
       onGoBack();
     }
+    onCloseProp?.();
```
RNR1: nit: For all these cases, can we follow the guard convention `onGoBack` uses?
```ts
if (onCloseProp) {
  onCloseProp();
}
```

Issue #768, opened 2026-09-02
Enable global Addresses


affabillyty:
**Is your feature request related to a problem? Please describe.**
As a user, when submitting a new resource, the "Address" field blocks me from submitting anything outside of the US + Canada.

**Describe the solution you'd like**
I would like the Location + Address features to enable submissions from anywhere in the world. This means that I could start typing in an address in the Address field and it would enable anything -- like Google Maps. It also means that using the "Use My Location Instead" button would work and help populate the address as needed. The Address field is both in the "Add a Site" form and the "Suggest an Edit" form.

**Describe alternatives you've considered**
N/A

**Additional context**
IMPORTANT: the same issue is present on the Search bar that is accessed from the toolbar. The Search bar should also enable global results.

PR #770, opened 2026-09-08
Enable global address autocomplete

icycoldveins:
# Pull Request

## Change Summary

- Remove the US-only region restriction from the shared Google Places autocomplete query
- Enable worldwide address results in Search, Add a Site, and Suggest an Edit
- Add a standalone global-address Cypress spec without changing the existing search tests

## Change Reason

Users need to be able to search for and submit resources at addresses outside the United States.

Related Issue: #768

https://github.com/user-attachments/assets/e9097c85-306f-4699-af36-93a7cfb71a65


RNR1 (approved):
Code LGTM, Something to note about this approach is that you will get results that are location biased only if you share your location (e.g., I'll only see results from Philadelphia first if I have enabled location and I'm in Philadelphia). That might has some UX tradeoffs, so I think it would be good to communicate it with the team before merging

PR #773, opened 2026-09-16
Update agent instructions to use the current package manager (pnpm)

### LINKED ISSUE #762: Update CLAUDE.md to reflect pnpm migration
**Describe the bug**
CLAUDE.md file still references yarn commands, but the repository has shifted to pnpm

**Screenshots**
Image

**Additional context**
- Update it to know the latest context and Dev Exp, this prevents the need of arguing with Claude 😂

nathanktruong:
# Pull Request

## Change Summary

Replaced references in the Claude instructions from `yarn` to `pnpm`

## Change Reason

The repo does not use `yarn`, so it's misleading to readers and those running its scripts.

## Verification [Optional]

I don't have an agent installed on my computer, so here's a screenshot of the file:

Screenshot 2026-09-15 at 8 27 46 PM

Related Issue: #762


nathanktruong:
question: The fetch supabase script is the only other area that references `yarn` for running the script. Should we fold that in here since it only affects comments?

Screenshot 2026-09-15 at 8 25 18 PM



Good find,

If changing this to pnpm works fine, feel free to edit that as well.

AnilKumar3494 (changes requested):
If the fetch function works fine with pnpm (which I believe should) update that as well

nathanktruong:
> If the fetch function works fine with pnpm (which I believe should) update that as well

Thanks Anil! Doesn't work. Looks like it wasn't carried over in the move from yarn. I'll create a follow-up ticket and leave a TODO comment.
Screenshot 2026-09-23 at 12 08 03 PM

nathanktruong:
Re-requested review. I updated that piece of code to document that its broken, and opened a new issue to investigate it. Won't be renaming it to `pnpm`.

